Issue #115, opened 2014-10-07
Interpolation markup


jeton:
Hey guys, 

Really nice job. Just a quick one, and probably edge-case. Have you guys thought about abstracting out the interpolation markup in the bootstrap decorator html files? I'm a python django user and can't use the default {{ }} syntax. 

I guess it's not a problem for me to change if I fork, edit the html files and re-gulp it.

mayaboy:
Hi All
Sorry for reopening this.
Could you explain to me how and where replace the {{ }} ?
I downloaded  the schema-form.js but I couldn't find any trace of {{ }} , except in the var defaultMessages 
Is it something I have to change in the dependencies files?
In my app, I'm using a custom interpolation using this approach:
.config(function($interpolateProvider) {
    $interpolateProvider.startSymbol('{[{');
    $interpolateProvider.endSymbol('}]}');
  })

Thanks for your help.

mayaboy:
I found it
I'm changing the bootstrap-decorator.js

vishnucr:
Hey, i have the same problem, i need to change the default interpolation. because i'm using python django. what should i do, ?

PR #116, opened 2014-10-07
Jeton's interpolation provider

jeton:
Relating to my previously stated issue (https://github.com/Textalk/angular-schema-form/issues/115) I have edited the template files, removing all references to hardcoded interpolation markup. I have also create a directive to be able to set a type and placeholder attrs of an input field without the need to for interpolation markup (default.html).


dmi3000:
What ever happened to this code? I can't seem to pull down any version that contains it.

davidlgj:
No it did not. Basically it's not the right solution either, not using expressions cripples us alot. And will be hard to enforce on third party add-ons. Especially when moving to bind once variants found in Angular 1.3. 

A better solution is to search and replace {{}} in the templates with something else and use the $interpolatipnProvider

zwiy:
Is there any new solution for this problem/issue? Or is it better to search and replace?

Anthropic:
@zwiy not aware of any other options sorry.

vishnucr:
how can i replace {{ }} interpolation

Issue #177, opened 2014-12-02
Set field dirty and not valid


jdgblinq:
I need to be able to set a field as $dirty = true and $setValidity(e,false), as per this guide: http://fdietz.github.io/recipes-with-angular-js/backend-integration-with-ruby-on-rails/validating-forms-server-side.html

How do I do that with angular-schema-form? I see how to set the form as $dirty, but how do I set an individual field?

Thanks

rhelgeby:
Exposing $$key$$ and setting the name attribute like you did would be useful by itself. Then, in the submit callback we will be able to call myForm['fieldName'].$setValidity(...).

davidlgj:
@jdgblinq I agree that model[key + '_error'] looks a bit hackish :-) I'll try to get time to check out the PR today or tomorrow

PR #178, opened 2014-12-03
Provide tools to allow user to set field dirty and not valid, providing

jdgblinq:
dynamic error message.

[#177]

To summarize:

1. Add name attribute to schema-form input fields. This allows us to query these fields by name, otherwise
    they're rather anonymous. For example, now we can do this in our controller:
    ```$scope.organizationEditForm[key].$dirty = true;```

    In the above example, ```key``` is a name attribute, and ```organizationEditForm``` represents a form. ```$dirty``` is used by schemaForm and Angular to indicate a form field with an error, so being able to set it is critical.
1. schemaForm now checks for the existence of model[key + '_error'], where key is the name of the field, as
    in 1, above. For example, if the field name is 'name', it checks for model['name_error']. If that exists, 
    schemaForm uses it's contents as the error message. This allows us to dynamically pass a field specific
    error message to schemaForm, all the way from the server side, which is a bit like the west side, 'cause it's
    just so far away.


qstrahl:
@jdgblinq FYI you need to rebase this PR

jdgblinq:
@qstrahl I don't really think there is a chance of it being accepted as-is.

davidlgj:
For anyone wondering this is being worked on. No promises on ETA though.

davidlgj:
Hi @jdgblinq, just wanted to say that I haven't forgotten about this issue.

I just refactored error messages (its in development) so while it doesn't fix this its the first step. I'm thinking about the api for this. I'm considering adding events that lets you target specific fields and set their validation status and $dirty. Possibly even injecting validation messages. 

Something like 
```javascript
scope.$broadcast(
  'shemaForm.user.name', 
  {
    error: 'backend', 
    validationMessage: 'Username {{value}} is already taken'
  }
)
```

What do you think?

jdgblinq:
I like it.

davidlgj:
Ok @jdgblinq, I added it to development, but I'm not sure its useful enough. 

So it works like this, you send and event with the key in the name, but it should start with `schemaForm.error.`. It takes an error code, an optional validationMessage and optional validity boolean (defaults to false).

This sets an error.
```js
scope.$broadcast('schemaForm.error.name', 'backendError', 'This is out town, scrub')
```

This clears it again.
```js
scope.$broadcast('schemaForm.error.name', 'backendError', true);
```

So that is my main issue with it right now, you have to manually clear it before posting otherwise the form will never validate. But that might be needed to keep flexibility in use cases... I'll think I'm going to let it rest for a while before documenting it. 

The commit is here https://github.com/Textalk/angular-schema-form/commit/8d89396c904a2c3a813aee2e9ba2b0cf484c544d

oatkiller:
:+1: Excited for this change. If I can help you out at all, please let me know!

davidlgj:
@oatkiller try out the development branch and find bugs :-) 

I'm hoping to do a release tomorrow afternoon (CET), but it all depends on how the testing will pan out.

davidlgj:
Closing this now since 0.8.0 is out. Thanks everyone!

PR #180, opened 2014-12-03
add showConditional expressions and conditional disabling

antoniopuero:
Added more complex logic for hiding or showing any element of the form. Also conditional disabling, not only through static readonly property.
But there're two lines which were unexpectedly added to the pr,
+              if (part.key && part.schema && angular.isDefined(part.schema['default'])) {
-                sfSelect(part.key, scope.model, part['default']);		
+                sfSelect(part.key, scope.model, part.schema['default']);
 and I can't figure out why they are there.

Also, it's kinda hard to write/read your spec :(


Dervisevic:
I think we'll have to give this another week to see if @antoniopuero responds or close it. It's been a month of additional information needed.

antoniopuero:
Hi, sorry for late response. I'm a bit unsure what do you mean by "use conditional with one read only field and one "ordinary"". List of dependencies helps us to keep conditional expression clear and simple, and in the future, when model may became smth more complex, we won't let user change the expression (for example implementation of visibility and category in form definition to be able to structure your model will let us to handle expression properly and this is not very robust). The other case is that user should not know how model holder is called inside of directive, I mean that without dependencies  list we should write "model.key1 === true". about changing all add-ons  you're right, don't know how to manage it better

dist/schema-form.js
```diff
@@ -226,6 +231,67 @@ angular.module('schemaForm').provider('schemaFormDecorators',
...
+
+              if (angular.isUndefined(expressionString)) {
+                return false;
+              }
+
+              var disabled = evalExpression(expressionString);
+
+              if (angular.isDefined(scope.form.required)) {
+                scope.form.required = !disabled;
+                scope.form.schema.required = !disabled;
```
davidlgj: Changing the actual schema is not something we can do, either it is required or it's not.

antoniopuero: Yes, I fill awkward for this part. But maybe you can give me an advice, how to handle better situation when field is required only if it is visible. Don't you?

davidlgj:
What I meant was that you can use two consecutive conditionals in the form definition, with the same "contition" except that one is the negation of the other, that way one of them is always shown and slap on a "readonly" on one of them. 

ex.
```js
[
  {
    type: 'conditional',
    condition: 'person.confirm',
    items: ['email']
  },
  {
    type: 'conditional',
    condition: '!person.confirm',
    items: [
      {
         key: 'email',
         readonly: true    
      }
    ]
  } 
]
```

This is a bit cumbersome though, as noted #234, and it probably needs a new option to remove the value from the model while it's hidden, see #204. So there is certainly room for improvement, but I believe this is the way to do it. I think a "condition" on every field could be a solution that would ease up the typing. 

Also another way to solve it, especially if you like to change the schema, is to just change the schema and form definition and do a redraw of the form, unless the form is huuuuge it re-renders fast enough so the user doesn't notice.

So I'm sorry but we can't merge this as it is.

dist/schema-form.js
davidlgj: This is kind of a restriction we have by using JSON Schema. It becomes a bit complicated when your validation scenario is hard to describe with a JSON Schema.

What you can do is to re-redraw the form by changing the schema or the form definition, but change the entire schema/form instance (i.e. use angular.copy), otherwise it won't trigger a redraw of the form. 

A bit more correct solution would be to use `oneOf`, `anyOf` etc and that way you might be able to describe the validation you need from within the schema, but sadly we lack support for those. 

Could you give some more specifics on what such a situation might be? There might be some other way to solve it.

PR #189, opened 2014-12-26
Fire event, when rendering of form is ready

mrijken:
With this event, we can do some extra things to the rendered form. My use case is executing $.material.init() from https://github.com/FezVrasta/bootstrap-material-design. I can not do that with postprocess, because I need a fully rendered form in my use case.

### DIFF
--- a/src/directives/schema-form.js
+++ b/src/directives/schema-form.js
@@ -117,7 +117,7 @@ angular.module('schemaForm')
               }
             });
           };
-          $(element[0]).triggerHandler("sf-render-finished");
+          scope.$emit("sf-render-finished");
         });
       }
     };


Dervisevic:
Why do you use a jQuery specific trigger? Feels very specific. To me it feels more natural to emit an angular event on the rootscope (https://docs.angularjs.org/api/ng/type/$rootScope.Scope#$emit) in this case. Then anyone can set up listeners, and in your case that listener can run that triggerEvent.

mrijken:
The usage of jquery has no special meaning; I have changed it to emit.

Dervisevic:
I've merged it into development and it'll come out with the next release.

PR #191, opened 2015-01-02
Update index.md

stramel:
Just noticed a few problems with the example code. :)


Dervisevic:
Great catch, however we would greatly appreciate if you could submit the pull request against the development branch instead of master and i'll merge it right away.

stramel:
Oh, sorry! I just did an edit on the file. I will change it to develop.

PR #209, opened 2015-01-12
Added accessibility classes to default inputs

stramel:
Using `sr-only` to hide the title instead of `ng-show`.
Added `aria-hidden` to the feedback.

Wasn't sure how to implement the `sr-only` feedback part so I left it out.
The documentation about the changes I made and the one I didn't are here:
http://getbootstrap.com/css/#callout-has-feedback-icon-positioning


stramel:
This is passing tests but doesn't seem to be working when I try to used the code in-place of the bower installed code.

davidlgj:
Thanks for the PR! Aria stuff is definetly appriciated. 

I was wondering about the changes to title, using `sr-only` means that screen readers always gets a title even if the form definition says "notitle". I guess thats what your aiming at?

stramel:
@davidlgj Yes, the title using `sr-only` means that the `input` will always have a title even using `notitle`. This allows the `input` to always have a `label` for screen readers and using `sr-only` displays off screen for any non-screen reader. Perhaps, if you wanted, we could change/add conditions for `notitle` and `hidetitle`?

davidlgj:
No thats fine :-) Just wanted to check that it was intended behavior.

davidlgj:
Thanks! :+1:

PR #210, opened 2015-01-12
Update gulp tasks to output both unminified and minified in one file.

stramel:


davidlgj:
:+1: Thanks!

Dervisevic:
When this is merged we should change the `main` property of bower.json to have the unminifed files listed there as that is recommended by bower: https://github.com/bower/bower.json-spec

stramel:
@Dervisevic :+1:

PR #228, opened 2015-01-23
Call ngModel.$setDirty() when deleting from array

### LINKED ISSUE #102: The form is still pristine after removing an array element
When I load a schema with default values, and I remove an element from an array, `ngform.$pristine` is still true.
in `deleteFromArray` you actually trigger validation. Wouldn't it be useful to call `ngform.$setDirty()` too ?

(https://github.com/Textalk/angular-schema-form/blob/master/dist/schema-form.js#L1035 and https://github.com/Textalk/angular-schema-form/blob/master/src/directives/array.js#L96)

qstrahl:
- Closes #102

### DIFF
--- a/dist/schema-form.js
+++ b/dist/schema-form.js
@@ -1087,6 +1087,7 @@ angular.module('schemaForm').directive('sfArray', ['sfSelect', 'schemaForm', 'sf
             if (scope.validateArray) {
               scope.validateArray();
             }
+            ngModel.$setDirty();
             return list;
           };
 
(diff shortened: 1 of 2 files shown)


davidlgj:
Oh nice! But I think this is broken in Angular 1.2, since I believe it lacks the `$setDirty` method on the `ngModelController`. At least I can't find it in neither the docs or in the source (quick search).

I merged it anyway into develop and just surrounded it with an if, see https://github.com/Textalk/angular-schema-form/commit/bc4503e6901719c8867efb1851b3baedfb42410a

qstrahl:
Perfect, thank you!

Issue #100, opened 2014-09-24
Selective overrides in form definitions


snawbel:
I have a large schema and I would like to set some of the `type: 'string'` to textareas. However, if I use both the `'*'` and modify a key to be `'textarea'` in my form definition that field will be duplicated.

Am I missing something? It would be nice to be able to use defaults for all form fields except the ones explicitly modified in the form definition.

iamjohnlong:
Would there be a need to add another attribute? I would think if you use `"*"` in the form definition and then if you have a form definition that maps to the original default it would just override it in its proper place.

PR #233, opened 2015-01-27
Feature/overwrite form definition

iamjohnlong:
This addresses #100


davidlgj:
Hi @iamjohnlong, and thanks for the PR!

I skimmed the code, but I didn't get *how* it was supposed to be used? Could you explain a bit? 

(Nitpicking here, but why did you change the formatting of the comments?)

iamjohnlong:
This PR isn't going to work as is. It's working from my use case but it needs some work. 

This is how it should be working...

say I have a schema like so 

```
var schema = {
    "type": "object",
    "title": "Comment",
    "properties": {
      "name": {
        "title": "Name",
        "type": "string"
      },
      "email": {
        "title": "Email",
        "type": "string",
        "pattern": "^\\S+@\\S+$",
        "description": "Email will be used for evil."
      },
      "comment": {
        "title": "Comment",
        "type": "string",
        "maxLength": 20,
        "validationMessage": "Don't be greedy!"
      }
    },
    "required": [
      "name",
      "email",
      "comment"
    ]
  }
```

and my form definition looked like:

```
  $scope.form = [
    '*',
    {
      "key": "email",
      "type": "textarea",
      "placeholder": "Make a comment"
    },
    {
      "type": "submit",
      "style": "btn-info",
      "title": "OK"
    }
  ];
```

The email definition would overwrite the the email in the `"*"` form and not be duplicated. 

This would be helpful when you have very large schemas and you only need to customize small portions of the form.

iamjohnlong:
Now this should be working for the use case I described above. The comments where formatted by my text editor. I can switch them back if it's an issue.

src/services/schema-form.js
```diff
@@ -299,18 +327,30 @@ angular.module('schemaForm').provider('schemaForm',
...
       //simple case, we have a "*", just put the stdForm there
       var idx = form.indexOf('*');
       if (idx !== -1) {
-        form  = form.slice(0, idx)
-                    .concat(stdForm.form)
-                    .concat(form.slice(idx + 1));
+        // If there are more items in the form definition let's overwrite matches
+        if (form.length >= 1) {
+          form = form.slice(0, idx).concat(form.slice(idx + 1));
+          angular.forEach(lookup, function(k, v) {
```
davidlgj: I'm not following why you do this? wouldn't it be easier to just loop over and extend the objects in the `lookup` map? They are the defaults from the schema and therefore the standard form.

davidlgj:
Hi agin @iamjohnlong,

I've been playing around with your code, and though I really like the functionality, I'm a bit worried about having the overrides inside the form definition and relying on `"*"`. It feels a bit unintuitive and complex for me, let me give you an example: 

A common scenario is to just add submit (and probably cancel) buttons at the end of a form, this you can do today with: 
```javascript
[
  "*",
 { "type": "submit", "title": "Save" }
]
```

It's also possible to wrap the star in a fieldset or some such:
```javascript
[
  {
     "type": "fieldset",
     "items": ["*"],
 }
 { "type": "submit", "title": "Save" }
]
```

So with that in mind adding overrides, while technically possible, feels a bit weird, at first glance this code does not do what you think it does:
```javascript
[
 "*",
 { "key": "email", "I'm an override!" }, 
 { "type": "submit", "title": "I'm not" }
]
```

Also it depends on you using the "*" for the overrides to kick in, which might not always be what you want, say you have a complex deeply nested schema for users in you system, and you like to display it but only parts of it? ex: this wouldn't work
```javascript
[
 "contact",
 "address",
 { "key": "contact.email", "I'd like to override but I can't..." }, 
]
```

Would it work for you with an extra attribute instead? Say lilke `sf-override`? 
ex: 
 ```javascript
scope.form = [
 "contact",
 "address"
]

scope.override = { "key": "contact.email", "Overriding form defaults" } 
```

I think that would be very easy to implement. Another idea I was tinkering with is to add overrides as an extra attribute, but also make it possible to add them to the form definition, but as a special list in the list just to mark them out (and skip the depedency on '*'), ex; 
```javascript
[
 "contact",
 "address",
 [
   { "key": "contact.email", "I'm an overide." }
 ] 
]
``` 
The preferred method would still be the extra attribute since that is clearer but if you really need to include it in the form def you can.

What do you think?

iamjohnlong:
@davidlgj Ok the extra attribute sounds good. I will try and get something cooked up this weekend. I'll call it `sf-override`

davidlgj:
ok cool! If you don't get time I can try and get it fixed asap. But my
basic idea was to check for the attribute right after the stdForm has been
created from the schema and extend the objects in lookup by matching keys.

fre 6 feb 2015 16:04 John Long <notifications@github.com> skrev:

iamjohnlong:
@davidlgj if you want to take a stab at it go for it. I probably wont be able to get to it until Sunday. If I don't see anything then I will work on it. Thanks man!

davidlgj:
@iamjohnlong I didn't get much time this weekend, but what I got started I chucked up in the feature/overrideit branch. Checkout the simple example (don't forget to minify first though). Its WIP and keys needs to be normalized, fix so it works on consecutive overrides, tests written, docs etc.

iamjohnlong:
We'll need to handle conditional items as well. 

example:
```
{
  "schema": {
    "type": "object",
    "title": "Comment",
    "properties": {
      "name":  {
        "title": "Name",
        "type": "string"
      },
      "email":  {
        "title": "Email",
        "type": "string",
        "pattern": "^\\S+@\\S+$",
        "description": "Email will be used for evil."
      },
      "comment": {
        "title": "Comment",
        "type": "string",
        "maxLength": 20,
        "validationMessage": "Don't be greedy!"
      },
      "hasFavorites": {
        "title": "Add Favorite Things",
        "type": "boolean",
        "default": false
      },
      "favorites": {
        "type": "object",
        "title": "Favorite Things",
        "properties": {
          "color": {
            "title": "Favorite Color",
            "type": "string",
            "validationMessage": "Whats your favorite color!",
            "required": true
          }
        }
      }
    },
    "required": ["name","email","comment"]
  },
  "form": [
    "*",
    {
      "type": "submit",
	    "style": "btn-info",
      "title": "OK"
    }
  ],
  "overrides": [
    {
      "key": "comment",
      "type": "textarea",
      "placeholder": "Make a comment"
    },
    {
      "key": "favorites.color",
      "placeholder": "Your favorite color"
    },
    {
      "type": "conditional",
      "condition": "modelData.hasFavorites",
      "items": [
        "favorites"
      ]
    }
  ]
}
```

I'm trying to figure out the best way to handle this...

davidlgj:
Oh that is a tricky thing! Would it work if we changed the overrides to be an object instead of an array?
Then we can have the object attribute be the key that we want to override and the value be an array of form objects, so then one form field could be switched out with several. 

Something like this:

```json
{
  "schema": {
    "type": "object",
    "title": "Comment",
    "properties": {
      "name":  {
        "title": "Name",
        "type": "string"
      },
      "email":  {
        "title": "Email",
        "type": "string",
        "pattern": "^\\S+@\\S+$",
        "description": "Email will be used for evil."
      },
      "comment": {
        "title": "Comment",
        "type": "string",
        "maxLength": 20,
        "validationMessage": "Don't be greedy!"
      },
      "hasFavorites": {
        "title": "Add Favorite Things",
        "type": "boolean",
        "default": false
      },
      "favorites": {
        "type": "object",
        "title": "Favorite Things",
        "properties": {
          "color": {
            "title": "Favorite Color",
            "type": "string",
            "validationMessage": "Whats your favorite color!",
            "required": true
          }
        }
      }
    },
    "required": ["name","email","comment"]
  },
  "form": [
    "*",
    {
      "type": "submit",
        "style": "btn-info",
      "title": "OK"
    }
  ],
  "overrides": {
    "comment": [{
      "key": "comment",
      "type": "textarea",
      "placeholder": "Make a comment"
    }],
    "favorites.color": [{
      "key": "favorites.color",
      "placeholder": "Your favorite color"
    }],
    "favorites": [{
      "type": "conditional",
      "condition": "modelData.hasFavorites",
      "items": [
        "favorites.color"
      ]
    }]
  }
}
```

This could work I guess, but you need to be stringent about writing the overrides since an override that just wrappes the standard and if you invoke the standard inside it it would just recurse on itself....

iamjohnlong:
@davidlgj I've got it working with my original pull request. It was easier for me to just use the `stdForm.form` because I already had my head wrapped around it. I can try and refactor it to use the lookup though.

davidlgj:
@iamjohnlong Does that mean that this PR is stale?

mkormendy:
Well 6 years later, and I could really have used this.

PR #244, opened 2015-02-09
error on deleteFromArray task

bygiro:
the ngModel is undefined


davidlgj:
Hi @bygiro and thanks for the PR. It's merged into development.

Issue #230, opened 2015-01-26
validation doesn't work for radios


melnikaite:
http://jsfiddle.net/mdmg0ad3/

davidlgj:
Yep, that's a bug.

kupuguy:
I think that a simple fix for this would be to include the attribute `ng-required="!$$value$$"` in the input field attributes.

davidlgj:
@kupuguy sadly it's not that simple :-( First of it should only be required if the schema says so, not otherwise. And secondly we'd like to use our own error messages, not the browsers. 

Slapping on the directive `schema-validate="form"` on the input does indeed validate and stop submit, but I'll have to write some custom code to get error messages in the right place.

melnikaite:
It seems cause of it is `ng-repeat` in the template
I faced the same issue in my custom plugin

orderofout:
It seems like this issue is resolved but not merged with the master branch? Any timeframe to this or how I can get the fix?

ulion:
I can not image it do not support radios type's required!!!

how could we fix this?

davidlgj:
Fixed in 0.8.0, finally.

PR #249, opened 2015-02-11
fix validation inside ng-repeat

melnikaite:
#230

### DIFF
--- a/src/directives/schema-validate.js
+++ b/src/directives/schema-validate.js
@@ -30,12 +30,25 @@ angular.module('schemaForm').directive('schemaValidate', ['sfValidator', 'sfSele
         });
       }
 
+      scope.$parent.dirtyNgModelsList = []
+      scope.markSiblingsDirty = function(ngModel) {
+        var inDirtyNgModelsList = (scope.$parent.dirtyNgModelsList.indexOf(ngModel.$name) != -1);
+        if (ngModel.$dirty || inDirtyNgModelsList) {
+          if (!inDirtyNgModelsList) {
+            scope.$parent.dirtyNgModelsList.push(ngModel.$name);
+          }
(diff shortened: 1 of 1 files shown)


davidlgj:
I've haven't had time to test it, just looked at the code, but do as I can see there is actually no check  that we are inside an ng-repeat? 

And relying on scope.$parent is a bit fragile, just introducing a directive in a template can break things. Hmmm... I don't know. I'll do some testing when I get the time.

melnikaite:
Thanks for reply.
I also thought about option for providing scope that should be used instead of just using parent, because in some templates people can use more than one loop level, but unfortunately I don't have time to implement it.
So currently we can merge it as temporary solution.

davidlgj:
Are there any other scenarios except the radios that you have found?

melnikaite:
I haven't checked other templates with loops. Saw the problem only in radios and in my custom plugin that also use ng-repeat.

davidlgj:
Hi @melnikaite I haven't done any testing yet, sorry. But today I finally got around fixing better validation messages and with that it was easy to fix the problems with radios type not validating. 

Basically the changes that have landed in the development branch today let's you use a the `schema-validate`directive on the each of the radios (it goes hand in hand with ngModel) and then project the error message from one of them to below the radios via a new `sf-message` directive. 

It would be super awesome if you have time to check it out and see if that solves it for your custom plugin as well. I've updated the "extending" docs a bit, but basically you add a `schema-validate` to each element with an `ngModel` on it and a `<div sf-messages="form.description">` where you want your error message to go.

Dervisevic:
Has it solved the problem @melnikaite?

melnikaite:
Yes, thanks a lot
It started working in my fiddle http://jsfiddle.net/mdmg0ad3/

PR #257, opened 2015-02-19
Add license information for LibreJS compatibility.

albsod:
These commits

1. add javascript sources and license information following the FSFs guidelines (https://www.gnu.org/software/librejs/free-your-javascript.html). This lets the users a) examine what code gets exectuted in their browser and b) learn what code is available under a free license, so as to facilitate copying. With the added sources and license information users of the GNU LibreJS add-on, which blocks non-free javascript by default, will also be able to display the website correctly (with the exception of Twitter's follow button which doesn't seem to be free).

2. change a few http URLs to https

3. fix indentation here and there.


albsod:
By the way, "dist/bootstrap-decorator.js" does not appear to be the source file for "dist/bootstrap-decorator.min". (In "jslicense.html" I say it is.)

Dervisevic:
https://github.com/Textalk/angular-schema-form/blob/master/src/directives/decorators/bootstrap/bootstrap-decorator.js should be the source

index.html
```diff
@@ -214,166 +228,156 @@
...
-            'type': 'string'
+  // @license-end
+  </script>
+
+  <script>
+  // @license magnet:?xt=urn:btih:d3d9a9a6595521f9666a5e94cc830dab83b65699&dn=expat.txt Expat
+  //
+  // Actual usage of angular schema form code if you're looking for it.
+  hljs.initHighlightingOnLoad();
+  angular.module('schemaForm').controller('PageCtrl', function($scope, $http) {
```
davidlgj: Why is all of the code changed? Or at least why is it in the diff? It makes it very hard to see if you changed any code or not.

davidlgj: Oh did you use some kind of code formatter? It looks like a lot of empty lines have disappeared, that might be the problem.

albsod: Oh, I did some indentation cleaning, but I realize now that I shouldn't have done that. Too many different changes makes it harder for you to see exactly what I did. Those lines above that are marked as deleted should still exist in the file.

PR #261, opened 2015-02-23
Added optiongroup capability

DennisHalmstad:
using the group key in titlemap will make use of the angular function to make opiongroups.

### DIFF
--- a/src/directives/decorators/bootstrap/select.html
+++ b/src/directives/decorators/bootstrap/select.html
@@ -9,7 +9,7 @@
           sf-changed="form"
           class="form-control {{form.fieldHtmlClass}}"
           schema-validate="form"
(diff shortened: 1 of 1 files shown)


src/directives/decorators/bootstrap/select.html
```diff
@@ -9,7 +9,7 @@
           sf-changed="form"
           class="form-control {{form.fieldHtmlClass}}"
           schema-validate="form"
-          ng-options="item.value as item.name for item in form.titleMap"
+          ng-options="item.value as item.name group by item.group for item in form.titleMap"
```
davidlgj: Nice tweak. I haven't had time to test it, but I guess it works even if item.group is `undefined`?

davidlgj:
Tested it, looks good. Thanks for the PR!
Merged into development.

PR #267, opened 2015-02-25
Fix the radio controls name conflict

ulion:
the radios/radios-inline/radio-buttons fields, may occur the input name conflict if in the same form multiple properties has same last level property name even their "key" is different.

This PR fixed this by use "key" as radio type's input name.


davidlgj:
Thanks for the PR!

Issue #273, opened 2015-02-27
[Feature Request] Field Addons / Inline Forms


quietlyconfident:
For our application, we'd like:

1. The ability to style the form and/or individual fields as bootstrap inline-forms, and
2. The ability to add bootstrap input-group-addons to the fields themselves ( just icons / glyphs / etc; not other form inputs)

I think that the easiest way to do this would be to:

1. Fork ASF,
2. Add an option ( inlineBootstrap ), defaulting to false, to the globalOptions, as well as to each of the existing fields,
3. If inlineBootstrap is set globally (for the whole form) and/or locally (for the field) the decorator widget should show an inline-version of the field rather than the standard.
4. Add two options (fieldAddonLeft and fieldAddonRight) to the existing fields,
5. If an add-on option is set, then the field is displayed inside of a bootstrap input group, and the value of the option is used as the contents of the input-group-addon span.

*Most importantly* if we built this, would the maintainer be receptive to merging the changes, assuming sufficient code quality?

And, any suggestions on a better way to do it?

quietlyconfident:
The goal BTW is to be able to attach icons to fields like so:
![screen shot 2015-02-27 at 3 57 24 pm](https://cloud.githubusercontent.com/assets/249930/6421272/18df7516-be9b-11e4-9e74-02826709313c.png)

Dervisevic:
I can't speak for @davidlgj but if the code is clean i can't see why we wouldn't merge it.

quietlyconfident:
See #275

davidlgj:
Same as @Dervisevic said :)

davidlgj:
@quietlyconfident  I'll try to merge it tomorrow... have you done any work on doing inline forms? It's an often requested feature and I was hoping to soon look into it.

quietlyconfident:
I've made a bunch of inline forms manually (e.g. just writing HTML with
bootstrap styles) but not with ASF.  Left to our own devices we will
probably implement it within the next month or two for our application and
I'd be happy to contribute it.

-e

plong0:
@quietlyconfident wondering if you've made any headway on the inline forms for ASF?
looking at implementing them as my next task...

quietlyconfident:
Hello-

No, I moved onto other projects and it isn't looking like I'll get to it
anytime soon.  Go for it!

-e

Issue #164, opened 2014-11-07
Missing Required Controller


kuldeepcode:
Commit 18f4226 breaks validation of a schema / form combination that works for me in the prior commit, with the following error message URL:

```
https://docs.angularjs.org/error/$compile/ctreq?p0=ngModel&p1=schemaValidate
```

Which displays the message: Controller 'ngModel', required by directive 'schemaValidate', can't be found!

Controller 'ngModel', required by directive 'schemaValidate', can't be found!

davidlgj:
What is the schema/form combination that fails?

javifr:
+1

quietlyconfident:
Sorry for not responding earlier and leaving an unhelpful issue without follow-up details.

The schema / form combination I was having trouble with made use of custom fields / directives / types that I had written and, when I re-wrote them, the issue went away, so it became not a problem for me.

I'm still not sure why this commit triggered the issue.  It looks like this commit fixed some other bug which was preventing exposure of my bug before?  Not sure.

Anyways I think it should be closed as when I fixed my custom directives the error went away.

quietlyconfident:
Hello @Dervisevic  , some more information and (potentially) a request to re-open, though unfortunately I don't have a solution yet.

I've figured out that this error is triggered by including an ng-if in a template used by a directive/decorator, and putting the input element inside that ng-if block, or even putting the ng-if block on the input element itself.  For example, lines x-y of the default.html template for the bootstrap directive are:

````
<input ng-show="form.key"
       type="{{form.type}}"
       step="any"
````

If you insert an ng-if into that, like so:

````
<input ng-show="form.key"
       type="{{form.type}}"
       ng-if="true"
       step="any"
````

the error occurs.  This makes it difficult to make (what would otherwise be) simple modifications to the default decorator templates

I don't have a solution yet but the following information may be helpful:

The schema-validate directive, on line 8, requires the ngModel controller like so:  `require: 'ngModel',`

According to the [angular developer guide](https://docs.angularjs.org/guide/directive), in the section "Creating Directives that Communicate:"

> The myPane directive has a require option with value ^myTabs. When a directive uses this option, $compile will throw an error unless the specified controller is found. The ^ prefix means that this directive searches for the controller on its parents (without the ^ prefix, the directive would look for the controller on just its own element).

[...]

PR #275, opened 2015-02-28
Add bootstrap input-group-addon with option fieldAddonLeft/Right

quietlyconfident:
Implements the second of the feature requests in #273

### DIFF
--- a/src/directives/decorators/bootstrap/default.html
+++ b/src/directives/decorators/bootstrap/default.html
@@ -1,20 +1,27 @@
 <div class="form-group schema-form-{{form.type}} {{form.htmlClass}}"
      ng-class="{'has-error': hasError(), 'has-success': hasSuccess(), 'has-feedback': form.feedback !== false }">
   <label class="control-label" ng-class="{'sr-only': !showTitle()}" for="{{form.key.slice(-1)[0]}}">{{form.title}}</label>
-
-  <input ng-show="form.key"
-         type="{{form.type}}"
-         step="any"
-         sf-changed="form"
-         placeholder="{{form.placeholder}}"
-         class="form-control {{form.fieldHtmlClass}}"
-         id="{{form.key.slice(-1)[0]}}"
-         ng-model-options="form.ngModelOptions"
-         ng-model="$$value$$"
-         ng-disabled="form.readonly"
-         schema-validate="form"
-         name="{{form.key.slice(-1)[0]}}"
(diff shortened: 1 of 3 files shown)


src/directives/decorators/bootstrap/default.html
```diff
@@ -1,20 +1,27 @@
...
-         placeholder="{{form.placeholder}}"
-         class="form-control {{form.fieldHtmlClass}}"
-         id="{{form.key.slice(-1)[0]}}"
-         ng-model-options="form.ngModelOptions"
-         ng-model="$$value$$"
-         ng-disabled="form.readonly"
-         schema-validate="form"
-         name="{{form.key.slice(-1)[0]}}"
-         aria-describedby="{{form.key.slice(-1)[0] + 'Status'}}">
+  <div ng-class="{'input-group': (form.fieldAddonLeft || form.fieldAddonRight)}">
```
davidlgj: I like your commit in general, but this introduces an extra div for everyone... I'm not sure its a good idea, but maybe we could try having two versions and switching between them with an ng-if?

Something like:
```
<div class="input-group" ng-if="(form.fieldAddonLeft || form.fieldAddonRight)}">
...
</div>
<input  ng-if="!(form.fieldAddonLeft || form.fieldAddonRight)}"  .... >
```
What do you think?

quietlyconfident:
The bug in question is #164 .  To not have the doc, we could go bigger, and do ng-if on the outermost div, but that's not very DRY.

Also I took a look at the failed test cases, I think it is not failing because of an actual bug. But rather because the implementation of the test cases is very tightly coupled to the implementation of the HTML; putting in the extra div layer you noted, which should have no other effect, breaks those test cases.

davidlgj:
Yeah I figured that the tests where breaking because of the extra div. I agree that the outermost div is not a good idea. 

The fear I have is mostly due to peoples CSS breaking, but with proper documentation/changelog that should be easily fixable. I have time this Friday to actually do some coding and not just answer issues, so I'll take a look at it then if that's OK with you?

quietlyconfident:
Sounds great!  Thanks for considering the patch.

davidlgj:
We'll time slipped away for me, I got a lot of schema form things done, but I haven't had time to merge this yet, but it will be done before next release.

davidlgj:
Merged into development, I found the fix for #164 so I took the liberty of rearranging it so we got rid of the extra div. That fixed the test as well. I need to document it though before there can be a release. 

Thanks again!

Issue #276, opened 2015-03-01
TabArray fails when title has spaces


Coridyn:
In `tabarray.html` if the `form.title` is a string containing spaces then the interpolation will fail.

I've made a demo showing the issue:
https://github.com/Coridyn/angular-schema-form/tree/demo/tabarray-broken
(Select the "Tab Array (Broken)" example)

and here is a schema+form that does the same thing:

```json
{
  "schema": {
    "type": "object",
    "title": "Comment",
    "properties": {
      "breaks": {
        "type": "array",
        "items": {
          "type": "object",
          "properties": {
            "name": {
              "title": "Name",
              "type": "string"
            }
          },
          "required": ["name"]
        }
      }
    }
  },
  
  "form": [
    {
      "type": "help",
      "helpvalue": "<h4>This tabarray breaks when the `title` has spaces:</p>"
    }, 
    {
      "key": "breaks",
      "type": "tabarray",
      "title": "My name",
      "items": [
        "breaks[].name"
      ]
    }
  ]
}
```

The `tabarray.html` template tries to interpolate `form.title` which will fail if it contains spaces (because it's not a valid Angular expression):

```html
<a href="#">{{evalExpr(form.title,{'$index':$index, value: item}) || $index}}</a>
```

davidlgj:
Thanks again!

PR #277, opened 2015-03-01
Add `interp()` view helper

Coridyn:
Addresses Textalk/angular-schema-form#276

Adds a new `interp()` view helper that will interpolate strings, optionally containing Angular expressions.

This is almost the same as `evalExpr()` but doesn't assume the content is an expression.
Instead it parses expressions that use the standard Angular escape sequence: `{{ }}`

This demo shows how this can be used to address #276 (with a custom `tabarray` field):
https://github.com/Coridyn/angular-schema-form/tree/demo/tabarray-broken
(Select the "Tab Array (Fixed)" example)

### DIFF
--- a/docs/extending.md
+++ b/docs/extending.md
@@ -76,6 +76,7 @@ available here for you to act on.
 | errorMessage(msg) | Error message formatting, makes validationMessage option work. |
 | evalInScope(expr, locals) | Eval supplied expression, ie scope.$eval |
 | evalExpr(expr, locals) | Eval an expression in the parent scope of the main `sf-schema` directive. |
+| interp(expr, locals) | Interpolate an expression which may or may not contain expression `{{ }}` sequences |
 | buttonClick($event, form)  | Use this with ng-click to execute form.onClick |
 
 ### The magic $$value$$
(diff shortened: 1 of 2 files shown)


davidlgj:
Hi @Coridyn, nice PR!  Love that you updated the docs as well. Is there a reason though for not updating the `tabarray.html` file to use the new `interp()`?

Coridyn:
You're welcome, I'm glad to help - I'm quite interested in this project so I might have a look at some of the other issues to see if I can lend a hand there.

Changing the default `tabarray.html` would be a breaking change if people are currently expecting `form.title` to be interpolated. 

The more conservative approach was to add the `interp()` helper and let people make the change in their templates if needed, so that's what I went with.

I think having `interp()` as the default behaviour for `tabarray.html` (and others if needed) would be good. If you're happy with that I can make the change and submit another PR.

davidlgj:
I didn't actually realise it was a breaking change, but IMHO that is not a big issue since it's already broken (as you pointed out in #276), we'll bump the the version to 0.8.0 and mention it in the docs and changelog so people can change their titles. Somewhere soon we probably should go to 1.0.0 and start using semver versions properly... 

So yes a PR with that would be awesome! And thanks again :+1:

davidlgj:
And oh, feel free to look at any issue. We are seriously swamped at the moment and any help we can get is very very welcome. Just mail me if you have any questions.

PR #281, opened 2015-03-02
Docs: Change 'decoratorsProvider' to 'schemaFormDecoratorsProvider'

Coridyn:
Just a small change, the documentation referred to `decoratorsProvider` when it should be `schemaFormDecoratorsProvider`.

It was a little confusing when trying out the decorators ;)


Dervisevic:
Brilliant!

PR #283, opened 2015-03-05
Add form validation for radios types and fieldset

### LINKED ISSUE #270: required Object field validation does not work
like #251 shows, it not works, but with a more reasonable sample here:

schma:
```
{
  "type": "object",
  "title": "Comment",
  "properties": {
    "name": {
      "title": "Name",
      "type": "string"
    },
    "expirationDate": {
      "type": "object",
      "properties": {
        "month": {
          "type": "string",
          "enum": ["Jan","Feb"]
        },
        "year": {
          "type": "string",
          "enum": ["2014","2015"]
        }
      },
      "minProperties": 1
    }
  },
  "required": [
    "name",
    "expirationDate"
  ]
}
```

form
```
[
  "name",
  "expirationDate",
  {
    "type": "submit",
    "style": "btn-info",
    "title": "OK"
  }
]
```

I'd like user at least input either month or year, but this validation does not works, because even the null/undefined object field passed the "required" check. any idea why and how we can fix this?

ulion:
Add form validation to radios types and fieldset (schema object type). This may partly resolve #270, #230. though still no whole schema validation apply to the model when submit.

### DIFF
--- a/src/directives/decorators/bootstrap/fieldset.html
+++ b/src/directives/decorators/bootstrap/fieldset.html
@@ -1,5 +1,6 @@
-<fieldset ng-disabled="form.readonly" class="schema-form-fieldset {{form.htmlClass}}">
+<fieldset ng-disabled="form.readonly" class="schema-form-fieldset {{form.htmlClass}}" schema-validate="form" ng-model="$$value$$"
+     ng-class="{'has-error': hasError(), 'has-success': hasSuccess()}">
   <legend ng-show="form.title">{{ form.title }}</legend>
-  <div class="help-block" ng-show="form.description" ng-bind-html="form.description"></div>
+  <div class="help-block" ng-show="(hasError() && errorMessage(schemaError())) || form.description" ng-bind-html="(hasError() && errorMessage(schemaError())) || form.description"></div>
   <sf-decorator ng-repeat="item in form.items" form="item"></sf-decorator>
 </fieldset>
--- a/src/directives/decorators/bootstrap/radio-buttons.html
+++ b/src/directives/decorators/bootstrap/radio-buttons.html
@@ -1,4 +1,4 @@
-<div class="form-group schema-form-radiobuttons {{form.htmlClass}}"
+<div class="form-group schema-form-radiobuttons {{form.htmlClass}}" schema-validate="form" ng-model="$$value$$"
      ng-class="{'has-error': hasError(), 'has-success': hasSuccess()}">
   <div>
     <label class="control-label" ng-show="showTitle()">{{form.title}}</label>
--- a/src/directives/decorators/bootstrap/radios-inline.html
+++ b/src/directives/decorators/bootstrap/radios-inline.html
@@ -1,4 +1,4 @@
-<div class="form-group schema-form-radios-inline {{form.htmlClass}}"
+<div class="form-group schema-form-radios-inline {{form.htmlClass}}" schema-validate="form" ng-model="$$value$$"
      ng-class="{'has-error': hasError(), 'has-success': hasSuccess()}">
   <label class="control-label" ng-show="showTitle()">{{form.title}}</label>
   <div>
--- a/src/directives/decorators/bootstrap/radios.html
+++ b/src/directives/decorators/bootstrap/radios.html
@@ -1,4 +1,4 @@
-<div class="form-group schema-form-radios {{form.htmlClass}}" ng-class="{'has-error': hasError(), 'has-success': hasSuccess()}">
+<div class="form-group schema-form-radios {{form.htmlClass}}" schema-validate="form" ng-model="$$value$$" ng-class="{'has-error': hasError(), 'has-success': hasSuccess()}">
   <label class="control-label" ng-show="showTitle()">{{form.title}}</label>
   <div class="radio" ng-repeat="item in form.titleMap" >
     <label>


davidlgj:
Hi @ulion,

and thanks for the PR!  I'm sorry to say that I'm not going to merge it. A fieldset should not have any validation on it, especially since you can choose what fields goes into it, which makes it super easy to break your form, for instance by having validating against required fields that are not even in the form. (this is also a backwards incompatible change).

If you need full object validation I suggest you you make a new field type in an add-on for it.

Whole schema validation is not a use case we're trying to fix since so much of the validation JSON schema does is not something that can be mapped onto a form. At least not in a way that is actually usable. Larger more complicated validation scenarios is much better suited to add-ons with custom field types so that you can build the UI for the form in a way that suits your specific use case, instead of a generic solution that while correct feels unintuitive for the user.

Regarding the radios I will hopefully have a fix out for that today.

ulion:
I don't think so, as fieldtype, it bound to a ng-model, which is just the
object in the schema which defined it. for user defined fieldset, it has no
"key", then no "model" will be bound, it's easy to tell it and skip
validation if it's the case.

it's rather strange you guys just see it lack the basic validation features
but do nothing, and even close directly without discussion, please, do
something to fix the validation.

many things can be done to make it works without affect existed use cases,
it's not the excuse to do nothing.

2015-03-06 19:41 GMT+08:00 David Jensen <notifications@github.com>:

> Hi @ulion <https://github.com/ulion>,
>
> and thanks for the PR! I'm sorry to say that I'm not going to merge it. A
> fieldset should not have any validation on it, especially since you can
> choose what fields goes into it, which makes it super easy to break your
> form, for instance by having validating against required fields that are
> not even in the form. (this is also a backwards incompatible change).
>
> If you need full object validation I suggest you you make a new field type
> in an add-on for it.
>
> Whole schema validation is not a use case we're trying to fix since so
> much of the validation JSON schema does is not something that can be mapped
> onto a form. At least not in a way that is actually usable. Larger more
> complicated validation scenarios is much better suited to add-ons with
> custom field types so that you can build the UI for the form in a way that
> suits your specific use case, instead of a generic solution that while
> correct feels unintuitive for the user.
>
> Regarding the radios I will hopefully have a fix out for that today.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Textalk/angular-schema-form/pull/283#issuecomment-77546602>
> .
>



-- 
Ulion

davidlgj:
Sure, checking the key should fix the backwards compatibility problem, but it still a very fragile field type that can land users with forms that will never validate. As I see it object validation is not a "basic validation feature" since an object does not directly translate to a suitable form *input* type, i.e. something you can actually change. 

What other "basic validation features" do you think we lack? 

There are lots of things we're working on and there are lots of things I like to fix, but at the end of the day I got to prioritize because my time is limited.

ulion:
basic of basic, radios type's required feature.

please re-read json-schema, since we use it, then we need respect the
validation feature of json-schema, and show the corresponding error
messages in proper level as possible.

I have no idea why you repeat consider the stupid use json schema but do
not want it be validated, form never passed? why pass the schema which
never validated there? it's the user's problem, he use the library in the
wrong way!! just simple adjust the schema to match the model that they want
to validate, then validate it.

2015-03-06 19:59 GMT+08:00 David Jensen <notifications@github.com>:

> Sure, checking the key should fix the backwards compatibility problem, but
> it still a very fragile field type that can land users with forms that will
> never validate. As I see it object validation is not a "basic validation
> feature" since an object does not directly translate to a suitable form
> *input* type, i.e. something you can actually change.
>
> What other "basic validation features" do you think we lack?
>
> There are lots of things we're working on and there are lots of things I
> like to fix, but at the end of the day I got to prioritize because my time
> is limited.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Textalk/angular-schema-form/pull/283#issuecomment-77548550>
> .
>



-- 
Ulion

davidlgj:
The radio type is just a bug, working on a fix now. Basically I had to refactor the way we do error messages so the validation message would so it took longer. 

But besides that bug?

> I have no idea why you repeat consider the stupid use json schema but do not want it be validated, 
> form never passed? why pass the schema which never validated there? it's the user's problem, he use 
> the library in the wrong way!! just simple adjust the schema to match the model that they want to 
> validate, then validate it.

I don't really follow you here, the first sentence does not make any sense. Do you mean that the form should always have exactly the same fields as in the schema? 

It's a *feature* and a design decision that you should be able to have one schema that validates your model but be able to have several form views for part of it, and that is one of the underlying reasons that we have a split between the JSON Schema and the form definition.

ulion:
no, the form not always have same field with schema, but the model should,
so the validation should always works.

for the multiple views to same schema, it's user's decision to split the
schema for each view to make it works and validate correctly even they
share same model. you should not insist to support what you do not support,
but do not support what you should support.

2015-03-06 20:30 GMT+08:00 David Jensen <notifications@github.com>:

> The radio type is just a bug, working on a fix now. Basically I had to
> refactor the way we do error messages so the validation message would so it
> took longer.
>
> But besides that bug?
>
> I have no idea why you repeat consider the stupid use json schema but do
> not want it be validated,
> form never passed? why pass the schema which never validated there? it's
> the user's problem, he use
> the library in the wrong way!! just simple adjust the schema to match the
> model that they want to
> validate, then validate it.
>
> I don't really follow you here, the first sentence does not make any
> sense. Do you mean that the form should always have exactly the same fields
> as in the schema?
>
> It's a *feature* and a design decision that you should be able to have
> one schema that validates your model but be able to have several form views
> for part of it, and that is one of the underlying reasons that we have a
> split between the JSON Schema and the form definition.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Textalk/angular-schema-form/pull/283#issuecomment-77551975>
> .
>



-- 
Ulion

davidlgj:
I disagree, If you need to validate the model against the entire schema that is up to you, its super easy to use tv4 to do it. And it's an entirely different use case. 

Schema Form generates forms with validation from a JSON Schema and a form definition, that form can validate **itself** against the schema. I'm sorry that you don't think its useful for you, but it's certainly what I need. 

I don't have much control over the schemas so I can't split or change them, but I do need to have several form that each individually validate *before* I go ahead and post it all to the server. 

I believe this adds flexibility that is much needed.

ulion:
you mis-understood what I try to said. the basic use case is, a schema, a
model, a form, and need validation, the very basic use case.

but if the schema include sub-level object type schema tree, then current
implement totally not works with that validation rules. that's the problem,
and very basic usages, e.g. one field value is required if the other field
is inputed, then use json-schema dependencies validations can do it, and
the error messages is put into the fieldset correctly with this PR. but
without it, can you tell me how to implement such validation? all form
input fields are generated dynamicly, totally no way to add such validation
and show error message in proper level.

2015-03-06 21:26 GMT+08:00 David Jensen <notifications@github.com>:

> I disagree, If you need to validate the model against the entire schema
> that is up to you, its super easy to use tv4 to do it. And it's an entirely
> different use case.
>
> Schema Form generates forms with validation from a JSON Schema and a form
> definition, that form can validate *itself* against the schema. I'm sorry
> that you don't think its useful for you, but it's certainly what I need.
>
> I don't have much control over the schemas so I can't split or change
> them, but I do need to have several form that each individually validate
> *before* I go ahead and post it all to the server.
>
> I believe this adds flexibility that is much needed.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Textalk/angular-schema-form/pull/283#issuecomment-77558019>
> .
>



-- 
Ulion

davidlgj:
As I said in the in the first comment, I don't want to add it to the standard fieldset, especially since you can break your own form. A functionality like this belongs in it's own field type, you can do it as an add-on or you can do a new PR (or update this if you want.) I'm not against that. 

I'm currently in the middle of some changes regarding error messages that I started working on this morning, I'll try to land it the develop branch today. I'd suggest you wait for that. Basically you should be able to just copy fieldset and rename it.

ulion:
shouldn't add a validation bool property to form definition? so the
validation can be switched on/off by the property?

2015-03-06 21:58 GMT+08:00 David Jensen <notifications@github.com>:

> As I said in the in the first comment, I don't want to add it to the
> standard fieldset, especially since you can break your own form. A
> functionality like this belongs in it's own field type, you can do it as an
> add-on or you can do a new PR (or update this if you want.) I'm not against
> that.
>
> I'm currently in the middle of some changes regarding error messages that
> I started working on this morning, I'll try to land it the develop branch
> today. I'd suggest you wait for that. Basically you should be able to just
> copy fieldset and rename it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Textalk/angular-schema-form/pull/283#issuecomment-77562084>
> .
>



-- 
Ulion

davidlgj:
It's nicer with a field type than an option in my opinion, makes it easier to document as well since it would need some documentation that points out that you should take care to include all the fields inside the validating fieldset that are required or in some other way can influence validation. 

I just pushed to development btw. The biggest change that might impact you in this is that there is now a directive , `sf-message`, that encapsulates and shows error messages. Which makes it easier to place it wherever you like. It's not tested enough yet so it it probably a bit rough.

ulion:
why should "include all the fields inside the validating fieldset that are
required"? it's user's duty to make the model validated to the schema he
supplied, if the model do not validate to the schema, then will show the
validation error message in the proper level of field/fieldset, then user
will realize that his model do not match schema, it is exactly the desired
behaviour.

2015-03-06 22:24 GMT+08:00 David Jensen <notifications@github.com>:

> It's nicer with a field type than an option in my opinion, makes it easier
> to document as well since it would need some documentation that points out
> that you should take care to include all the fields inside the validating
> fieldset that are required or in some other way can influence validation.
>
> I just pushed to development btw. The biggest change that might impact you
> in this is that there is now a directive , sf-message, that encapsulates
> and shows error messages. Which makes it easier to place it wherever you
> like. It's not tested enough yet so it it probably a bit rough.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Textalk/angular-schema-form/pull/283#issuecomment-77565993>
> .
>



-- 
Ulion

davidlgj:
I mean the end user, not the developer. Let's say you dynamically build form definitions from options you have in a database, you have lots of options and it's quite complex. You could introduce a bug that generates forms that  the end user never can fill in. Which is not a good situation. 

btw, I broke something in my last push to development. It's all broken now. Got to fix it.

ulion:
I don't think it's scope this library should considered, the library serve
developers, developers consider the problem you considered here to use it
correctly.

besides, I doubt the use case you worried is any common case. the normal
case is the developer generate the schema, and make it validate to the
model, the library should just do developer setup it to do, that's to
validate the model to the schema, then if error, if the error refer to a
certain node/level of field/fieldset, try to show the error message to
correct place, if not found, show error message in the form level if
possible. and supply the possibility to let developer add custom
validation, and the error message shows to certain field/fieldset level
(which currently no way imo, that's why I'm try to make the json-schema
validation to do it)

2015-03-06 22:38 GMT+08:00 David Jensen <notifications@github.com>:

> I mean the end user, not the developer. Let's say you dynamically build
> form definitions from options you have in a database, you have lots of
> options and it's quite complex. You could introduce a bug that generates
> forms that the end user never can fill in. Which is not a good situation.
>
> btw, I broke something in my last push to development. It's all broken
> now. Got to fix it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Textalk/angular-schema-form/pull/283#issuecomment-77568318>
> .
>



-- 
Ulion

davidlgj:
I think a framework should try to be easy to use. 

And I don't now how people use it, but its definitely our case. And I don't think yours is that common either, and as I said earlier I believe that in most cases you need to build a specific field type to get a good user friendly form when you have advanced validation (object level validation being that).

ulion:
that sounds really wrong.

one of the the main problem is angular-schema-form is hard to custom, every
form element's html was generated by hard coded template, while angular
normally use directive to extend function, then the result is funny, people
can not add directive to the dynamic generated form html. yes you said,
extend your form type to do that, really? it it the angular way to do that?
totally not. angular way should easily extend and reuse current templates,
not to rewrite or copy and rewrite.

the other main problem is validation. couldn't you validate the whole
schema to the whole model, then get the error message with the key/path
info, then use each key/path info to locate each related field/fieldset's
div, then let the error message shows in the correct place? this is what it
should do, not ask user to resolve it. as an user, I have totally no idea
how to locate to certain field/fieldset by the key path. do you?

2015-03-06 23:03 GMT+08:00 David Jensen <notifications@github.com>:

> I think a framework should try to be easy to use.
>
> And I don't now how people use it, but its definitely our case. And I
> don't think yours is that common either, and as I said earlier I believe
> that in most cases you need to build a specific field type to get a good
> user friendly form when you have advanced validation (object level
> validation being that).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Textalk/angular-schema-form/pull/283#issuecomment-77572531>
> .
>



-- 
Ulion

davidlgj:
I get it that "whole schema validation" seems to be the thing that you are stuck at. But it is apparent that we don't see eye to eye on this and that you ignore what I say and just harp on with the same thing. 

Schema Form is library that takes a JSON Schema and a form definition and generates a form *whose  fields get their validation from the schema*. This is by design and just the way we want it. 

I'm tired of arguing that with you, just because you yell the loudest doesn't mean you can have all of my attention. If you don't find schema form suits you then use another form generating library instead, there are loads of 'em. Or fork and fix what you perceive is broken. 

So good luck and I hope you find a solution to your use case.

orderofout:
Is there an update to the bug that does not validate a radio button or checkbox?
I'm not sure what you are all talking about in this thread but it seems to be a completely different issue. Checking to make sure a required checkbox or radio button is checked is a very valid issue and should be addressed.

davidlgj:
@orderofout It's probably resolved with the changes to validation messages that are in the development branch and will be out with next release. If you can please try it out. Any testing of the development branch right now is very appreciated, if you wan't too speed up the release cycle.

PR #288, opened 2015-03-09
Schema validation fix for Angular 1.2  ($setDirty)

adamcbuckley:
I am required to support IE8 and am therefore limited to Angular JS 1.2

I am sure you would make this fix in due course, but I have deadlines!  :-)

The Angular 1.2 version of $setDirty taken from http://makandracards.com/makandra/29625-setdirty-for-angular-1-2

I also fixed a typo - `if(ngModel.$setDirty())` should be `if(ngModel.$setDirty)`  (right?)


davidlgj:
:D Oh thanks, what a typo!

PR #290, opened 2015-03-09
Added method to set schemaError explicitly, in this case its possible to...

burdiuz:
Hi!
Added method to set schemaError explicitly, in this case its possible to do external schema validation and display proper notifications.
So, my problem is, I do additional validation using tv4 and my schema to enable complex validation cases with anyOf/oneOf/allOf. But I cannot set error to display proper validation message because error object is in closure. I've tried this in my project with making "error" property to be accessible via scope.error(and I've removed schemaError() method) but this seems not proper solution, so I've created scope.setSchemaError() method to setup my own error object.

### DIFF
--- a/src/directives/array.js
+++ b/src/directives/array.js
@@ -202,6 +202,9 @@ angular.module('schemaForm').directive('sfArray', ['sfSelect', 'schemaForm', 'sf
           // If there is a ngModel present we need to validate when asked.
           if (ngModel) {
             var error;
+            function setSchemaError(value) {
+              error = value;
+            }
 
             scope.validateArray = function() {
               // The actual content of the array is validated by each field
@@ -230,7 +233,7 @@ angular.module('schemaForm').directive('sfArray', ['sfSelect', 'schemaForm', 'sf
                 // Set viewValue to trigger $dirty on field. If someone knows a
                 // a better way to do it please tell.
                 ngModel.$setViewValue(scope.modelArray);
-                error = result.error;
+                setSchemaError(result.error);
                 ngModel.$setValidity('tv4-' + result.error.code, false);
               }
             };
@@ -244,6 +247,8 @@ angular.module('schemaForm').directive('sfArray', ['sfSelect', 'schemaForm', 'sf
             scope.hasError = function() {
               return ngModel.$invalid;
             };
+
+            scope.setSchemaError = setSchemaError;
 
             scope.schemaError = function() {
               return error;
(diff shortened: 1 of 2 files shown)


davidlgj:
Hi @burdiuz,

There are some undocumented changes in develop that also let's you set error and possible validation message on a field via events, see [last comment on pr 178](https://github.com/Textalk/angular-schema-form/pull/178#issuecomment-77644009)

Would you mind reading that and see if that would fix your use case? (I suspect it won't, but anyhow.) The PR is safe so I'll merge it if you need it, but it won't work with the new `sf-message` directive, but I guess you use this in a custom field type add-on?   

The ngModelController is also in scope inside a field type template so to push a error message into the `sf-message` directive you need to use `ngModel.$setValidity`  (for standard tv4 errors prefix error code with 'tv4-', i.e. 'tv4-302' for required) and add a matching [validationMessage](https://github.com/Textalk/angular-schema-form/blob/development/docs/index.md#validation-messages) if you don't like the standard message. 

This has the added bonus of actually setting the field to invalid, not just setting the error message.
So from a directive inside a custom field type with custom validation you could do like this.
```js
// push anyOf error
scope.ngModel.$setValidity('tv4-10', false);
```

As a side note we'd like to support anyOf,oneOf in the future, but I'm unsure on how, what is your use case? I.e. how does the form look like that uses these for validation?

burdiuz:
Hi!
Thanks for such fast answer, this event will work I guess. I've tried your solution with setting `$setValidity('tv4-'+error.code, false)` but the problem is -- I cannot get error message which came with error, so instead of message for `STRING_LENGTH_SHORT: 200` its always "Not a number". 
I am looking to show error message for error that was received from tv4 regardless of field type, but outside of schema-form flow. I'll try events you described and comment here.

davidlgj:
Did you try it against development branch? These changes aren't out in a release yet and it sounds awfully like what you would end up if you use master.

burdiuz:
Yeah, just tried and setting $setValidity() with development branch and it does set proper message if pass 'tv4-' type. Definitely much cleaner solution, thanks. When do you plan to release it?

My use case for `anyOf` is I am trying to make requirement of First Name + Last Name or Company in address form. So if Company name is filled, user is not required to fill First and Last Name fields and vice versa. There will be more forms with requirements like "at least one phone field must be valid", also I hope to use JSON Schema references.

FYI, it looks like developer branch has runtime error that was fixed in previous pull request
```
// FIXME: Try to make a fix for angular 1.2
        if (ngModel.$setDirty()) {
          ngModel.$setDirty();
        }
```

davidlgj:
Great! I'll close this PR then.

The plan is to do a release soonish, at the most on friday the 20th, since I know I got time to do it then. It's not that doing a release takes time, but I've done so much changes and I didn't have time to neither test manually nor write tests for the new events. Hope that works for you!

PR #296, opened 2015-03-12
When using the ui-select plugin there is a bug which likely affects all ...

strets123:
...array-based fields.

 If you add a multiselect without default value, even you can select some items on the
form but they are not captured in the model.

See the bug raised here: https://github.com/chengz/schema-form-uiselect/issues/2 and here https://github.com/networknt/angular-schema-form-ui-select/issues/3

I have fixed this in the default check cycle by adding an addtional condition that updates the underlying model to include any missing array fields.

Behaviour for ui-select is now the same as for the non schema form version.


strets123:
Apologies, just realised that this needs to be done in the individual plugins as it is already fixed centrally

PR #311, opened 2015-03-20
Add js license info for example page.

albsod:
This is a follow-up to #257.

### DIFF
--- a/examples/bootstrap-example.html
+++ b/examples/bootstrap-example.html
@@ -62,9 +62,16 @@
            ng-class="{red: !itParses}" ng-model="schemaJson" class="form-control schema"></div>
     </div>
   </div>
+  <div>
+      <p>
+          <small>
+              <a href="../jslicense.html" rel="jslicense">JavaScript license information</a>
+          </small>
+      </p>
+  </div>
 </div>
-<script type="text/javascript" src="http://code.jquery.com/jquery-2.1.1.min.js"></script>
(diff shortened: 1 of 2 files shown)


examples/bootstrap-example.html
```diff
@@ -62,9 +62,16 @@
...
+      <p>
+          <small>
+              <a href="../jslicense.html" rel="jslicense">JavaScript license information</a>
+          </small>
+      </p>
+  </div>
 </div>
-<script type="text/javascript" src="http://code.jquery.com/jquery-2.1.1.min.js"></script>
-<script type="text/javascript" src="http://cdn.jsdelivr.net/g/jquery.ui@1.10"></script>
+<script type="text/javascript" src="https://code.jquery.com/jquery-2.1.1.min.js"></script>
```
Dervisevic: Is there a reason you changed these to https? https can cause up to 5x slower load, and i personally don't see a reason that one would need to load these over https, its jquery.

albsod:
Personally, I always try to use HTTPS when possible. For me it's like turning the lights off when leaving a room. I wasn't aware of the speed difference though and I suppose there is no harm in using HTTP for the jquery URLs.

"I can see a case for using :// instead as that would remove it being blocked when going to browsing the https:// site."

OK, I'll change all the URLs to that.

Dervisevic:
Sorry it's actually // not ://

PR #313, opened 2015-03-23
Add global form option to disable tv4 validation

maff:
Useful in cases where you want to set validation messages only manually (e.g. from an API response) and use `angular-schema-form` primarily to render forms. When this config option is set on in `sf-options` it will completely omit tv4 validation for the form. Example:

```
<form sf-schema="schema" sf-form="form" sf-model="model" sf-options="sf-options="{ tv4Validation: false }"></form>
```

### DIFF
--- a/src/directives/schema-validate.js
+++ b/src/directives/schema-validate.js
@@ -41,6 +41,11 @@ angular.module('schemaForm').directive('schemaValidate', ['sfValidator', 'sfSele
           return viewValue;
         }
 
+        // Omit TV4 validation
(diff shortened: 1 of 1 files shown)


davidlgj:
Cool, this is a use case that never crossed my mind :)  Do you still have a need for a schema? Would it be a good thing to be able to use it *without* a JSON Schema?

maff:
For my use case I'm fine with JSON schema as I can generate all needed information on the backend.

PR #315, opened 2015-03-24
Fix sending tv4 standard message with broadcasted backend validation err...

mqamhieh:
In documentation it is mentioned that you can broadcast tv4- standard error messages in backend validation, but this is will not work as the code runs the validation just after receiving the broadcasted error. [Link to documentation] (https://github.com/Textalk/angular-schema-form/blob/development/docs/index.md#inject-errors-into-form-aka-backend-validation)


davidlgj:
Oh, good catch. Thanks!

PR #318, opened 2015-03-25
Update css bootstrap links in examples

Minisai:


Dervisevic:
That's good, i would prefer it if you kept the // instead of https:// as getting the files can take up to 5x as long, and that's not necessary when surfing without SSL. If you disagree, please do tell.

Minisai:
@Dervisevic Thank you for the remark! I've updated the PR. Protocol-related urls is good solution, but if you open file locally, browser would prefix it with "file://" protocol. As I understand these examples were created mostly for use on [github pages](http://textalk.github.io/angular-schema-form/examples/bootstrap-example.html), but at the same time I haven't found [custom-validators page](https://github.com/Textalk/angular-schema-form/blob/development/examples/custom-validators.html) there?

stramel:
@Minisai It looks like the custom-validators page was just created within the past week. I'm sure @davidlgj or @Dervisevic will get it up on the website soon.

I agree for protocol-less URLs :+1:

davidlgj:
@Minisai we should probably add some server to the gulpfile to make it easier to test out the examples. I usually use `http-server` or `puer`. The examples also play a second role to ease manual testing.

Issue #243, opened 2015-02-07
How to set form pristine after previous form data loaded?


ulion:
form.$setPristine() do not work, since it only setup $pristine on all inputs, but angular-schema-form use some decorator scope to check the ngModel's $pristine status.

I guess either because the scope, or the way form.$setPristine() does not set $pristine of the model caused the problem. But how to resolve it?

I do not want the green border when the form loaded with existed model data. how to do that?

ulion:
found the reason is at the scope.hasSuccess function of src/directives/schema-validate.js, which currently is:
```
scope.hasSuccess = function() {
    return ngModel.$valid && (!ngModel.$pristine || !ngModel.$isEmpty(ngModel.$modelValue));
};
```

all non-empty value forced be has success marker, which maybe not what we want when load existed model data.

rexebin:
one way to get ride of the green border would be go to bootstrap-decorator, and alter the default template: <div class="form-group" ng-class="{'has-error': hasError(),  'has-success': hasSuccess(), 'has-feedback': form.feedback !== false}">, get ride of the 'has-success': hasSuccess(),

It would be lovely if we could overwrite the template so we do not have to worry about updating and lose what we have changed in the code.

ulion:
But I'd like to change the hasSuccess() behavior with some way, if you can
supply some way to adapt it to do not consider non-empty as same as dirty
with some model level or controller level setting or tag?

2015-02-10 2:43 GMT+08:00 rexebin <notifications@github.com>:

> one way to get ride of the green border would be go to
> bootstrap-decorator, and alter the default template:
> , get ride of the 'has-success': hasSuccess(),
>
> It would be lovely if we could overwrite the template so we do not have to
> worry about updating and lose what we have changed in the code.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Textalk/angular-schema-form/issues/243#issuecomment-73564600>
> .
>



-- 
Ulion

rexebin:
hasSuccess() open add and remove the bootstrap css class "has-success" to the element, that is all the behavior. 

If I was to change the look of the green box, I would just simply overwrite bootstrap css class "has-success".

ulion:
No, I do want hasSuccess, but I want it only for dirty fields, what I want is:
```
scope.hasSuccess = function() {
    return ngModel.$valid && !ngModel.$pristine);
};
```

OR
```
scope.hasSuccess = function() {
[...]

PR #333, opened 2015-03-31
don't show has-success state until a field is dirty

cjroth:
This modifies the hasSuccess behavior to return true only when the field is dirty. See https://github.com/Textalk/angular-schema-form/issues/243#issuecomment-88175413

It would makes sense to add an option to show the hasSuccess when pristine too, if needed.


davidlgj:
I can't merge this as it is since this breaks current behavior that at least we rely on.  But I do see the use case. What we need is for it to be optional. Probably with a global option and control per field level as well.

cjroth:
Cool, that makes total sense. I'll add some options too.

davidlgj:
Awesome :+1:

shawnss:
Did this option get documented somewhere? I would like to use this as I do not want hasSuccess when form is rendered with existing data.

Dervisevic:
@shawnss It isn't in there yet it seems. @cjroth Do you plan on adding the options?

davidlgj:
@shawnss  I've started some work that incorporates this functionality, but as an option. https://github.com/Textalk/angular-schema-form/tree/hotfix/0.8.9

davidlgj:
@shawnss @cjroth  I've added an new global option `pristine` that supports this. Hope that works for you.

PR #351, opened 2015-04-13
Modern Javascript need quotes

rusxakep:


Dervisevic:
Interesting, do you have a source for this best practice?

davidlgj:
Why? And Is there a reason you went with " and not ' as per the style guide? (i.e. jscs set to 'google')

Dervisevic:
@rusxakep ?

PR #353, opened 2015-04-14
disableErrorState, disableSuccessState form options. fixes #214

### LINKED ISSUE #214: Disable input feedback
There is an option to turn off the icon feedback using `feedback: false` but it does not turn off the highlighting of the field ie. `input`. I believe `hasSuccess` and `hasError` should respect this flag as well.

domasx2:
Hi,

I made changes to field decorator ng-class as discussed in the issue. 
Was unable to find a way to mark checkboxes field as invalid though


davidlgj:
:star: :heart:  @domasx2 really nice!

PR #359, opened 2015-04-16
Please allow access to the current modelValue inside a condition

dben:
Allowing access to the value directly would make it easier to handle conditions off of nested arrays and dynamic schemas. 

I've included a pull request that demonstrates one way of allowing access to the current modelValue from a condition.


davidlgj:
Good idea @dben!  I patched it a bit so it uses ObjectPath for key stringification and it's only available when the form actually has a key.

PR #366, opened 2015-04-22
Modify conditional example

sangel10:
The conditional example wasn't working, the proposed solution works.


Dervisevic:
Thank you!

PR #370, opened 2015-04-27
Checkboxes decorator had misspelled schema-validate directive. Fixed typo.

jbsaff:
I was trying to apply some cleanup magic, triggering off of the presence of schema-validate. I thought I was going crazy because checkboxes weren't working, until I realized that they had a typo and thus didn't match the schema-validate directive.


jbsaff:
38f1553 is the only commit that applies to this PR. Sorry, I accidentally committed some other code on the same branch.

Dervisevic:
Oh wow, that's [embarrassing](http://i.imgur.com/wyEZ4.gif). Good catch man!

Issue #344, opened 2015-04-07
Config option to wipe the model when a condition is no longer fulfilled?


jbsaff:
I'd like wipe out user input from the model when the condition on a form field is no longer met. Currently the model is updated when a condition is met, but not when it no longer applies. Is there an option that I'm missing? A plain angular ng-if seems to do what I want, but that's because the form only submits elements in the DOM. I'd like to just take the schema-form model as my finished form response and go from there.

Any pointers would be greatly appreciated. Thanks!

davidlgj:
There has been some discussion on this over at https://github.com/Textalk/angular-schema-form/issues/204, so check that out. 

But he TL;DR; is that it's a good idea, at least to have it optionally.

jbsaff:
That's exactly what I'm thinking of. Thanks for the pointer. Sorry I missed the open discussion.

jbsaff:
I'm not a JS or AngularJS guy by trade (yet!), but I wanted to take a stab at solving my own problem. I had a couple of other form types/decorators I wanted, so I slipped in an extension to the schema-validate directive. https://github.com/jbsaff/angular-schema-form-multiselect for anyone interested.  It's not configurable yet, but if I make progress in that direction I'll submit a PR.

jbsaff:
Improved my solution and submitted it as a new feature via #371. I'm closing this issue because my initial question was answered and any follow-up can occur in the PR thread.

Issue #204, opened 2015-01-08
Should conditional type evaluation destroy data?


Anthropic:
I have an attribute directive I applied to my own decorator to remove data that was in elements hidden by a conditional type evaluation.

Should this perhaps be included in the main angular-schema-form behaviour?

```javascript
(function(window, angular, undefined) {'use strict';
  angular
    .module('schemaForm.extension',[])
    .directive('destroyHiddenData', ['oySchemaFormUtil', function(oySchemaFormUtil) {
      return {
        link: function(scope, element, attrs) {
          scope.$on('$destroy', function(){
            if(scope.form.key) oySchemaFormUtil.setKeyValue(scope.model, scope.form.key, '');
          })
        }
      };
    }]);
})(window, window.angular);
```
Note: the oySchemaFormUtil is my own function that just manipulates data based on key, but unlike the built in one it also allows for total deletion which is why I use it instead. Not relevant to discuss in this post thread.

davidlgj:
Yes, this might be a good idea... It would make conditional even more useful.

Maybe an option to turn it off if needed?

Anthropic:
I would suggest that the option allow for setting the value to a user preferred ( null, undefined, empty string or the current behaviour of do nothing) value, as in many cases it completely depends on how they then save the data to their database.

Anthropic:
Upon reflection I would like the default behaviour to set the value to null. It should not delete it as that is not ideal in JavaScript and the back end will need to be able to see changes without relying on every field always having a value.

jbsaff:
@Anthropic  I took a stab at this feature in https://github.com/Textalk/angular-schema-form/issues/371. I opted to use undefined instead of null, as that would most closely represent the state of how the user experienced the form (they don't see the question, so it never should come up in the model). However, I did include a check for a formDefaults definition of the destroyStrategy, which would allow for the entire form to use null as the mechanism for destroying data that has been hidden.

Anthropic:
@jbsaff great, I will keep an eye on it :)

Anthropic:
Completed by #371

PR #371, opened 2015-04-28
Feature: clean model when a field is removed from view, such as when a condition is no longer satisfied.

jbsaff:
This feature extends the schema-validate directive to allow users to configure an individual field or the entire form to one of a predefined set of behaviors when the $destroy event is broadcast. The common use case would be when a field is no longer visible due to a condition, which is evaluated by ng-if, no longer being met. Data that had previously been collected was remaining in the model, which would provide erroneous results on submission. By default, if no configuration is done, this feature would set the model value to undefined, just as if the field had never been visible in the form at all.

In order to properly handle complex nested arrays and objects, a new service (sfUnselect) was created. Based on the sfSelect service, sfUnselect will traverse a given path and attempt to set the endpoint to the provided value. If at any point an expected object or array is undefined, the service returns and does not create the missing object or array. 

This addresses both https://github.com/Textalk/angular-schema-form/issues/344 and https://github.com/Textalk/angular-schema-form/issues/204

Also, in case it comes up in testing, this is related to https://github.com/Textalk/angular-schema-form/pull/370, as the checkboxes will not work with this new feature until the schema-validate directive is correctly spelled.

Please pardon the commit log; I managed to fumble my git-fu roll when trying to create two pull requests.

### DIFF
--- a/src/directives/schema-validate.js
+++ b/src/directives/schema-validate.js
@@ -150,12 +150,10 @@ angular.module('schemaForm').directive('schemaValidate', ['sfValidator', 'sfSele
           if (destroyStrategy === 'retain') {
             return; // Valid option to avoid destroying data in the model.
           }
-          console.log('result %s', destroyStrategy);
 
           destroyUsingStrategy(destroyStrategy);
 
           function destroyUsingStrategy(strategy) {
-            console.log('Destroy called with %s', strategy);
             var strategyIsDefined = (strategy === null || strategy === '' || strategy === undefined);
             if (!strategyIsDefined){
               strategy = DEFAULT_DESTROY_STRATEGY;
--- a/src/directives/schema-validate.js
+++ b/src/directives/schema-validate.js
@@ -1,5 +1,5 @@
-angular.module('schemaForm').directive('schemaValidate', ['sfValidator', 'sfSelect', 'sfUnselect',
-  function(sfValidator, sfSelect, sfUnselect) {
+angular.module('schemaForm').directive('schemaValidate', ['sfValidator', 'sfSelect', 'sfUnselect', '$parse',
+  function(sfValidator, sfSelect, sfUnselect, $parse) {
 
     return {
       restrict: 'A',
@@ -128,48 +128,53 @@ angular.module('schemaForm').directive('schemaValidate', ['sfValidator', 'sfSele
         // Default behavior can be supplied as a globalOption, and behavior can be overridden in the form definition.
         scope.$on('$destroy', function() {
           var form = getForm();
+          var conditionResult = $parse(form.condition);
+          console.log(conditionResult(scope));
 
-          // Either set in form definition, or as part of globalOptions.
-          var destroyStrategy =
-              !form.hasOwnProperty('destroyStrategy') ? DEFAULT_DESTROY_STRATEGY : form.destroyStrategy;
-          var schemaType = getSchemaType();
+          if (form.hasOwnProperty('condition') && !conditionResult(scope)) { // If condition is defined and not satisfied.
 
-          if (destroyStrategy && destroyStrategy !== 'retain' ) {
-            // Don't recognize the strategy, so give a warning.
-            console.warn('%s has defined unrecognized destroyStrategy: \'%s\'. Used default instead.',
-                attrs.name, destroyStrategy);
-            destroyStrategy = DEFAULT_DESTROY_STRATEGY;
-          }
-          else if (schemaType !== 'string' && destroyStrategy === '') {
-            // Only 'string' type fields can have an empty string value as a valid option.
-            console.warn('%s attempted to use empty string destroyStrategy on non-string form type. ' +
-                'Used default instead.', attrs.name);
-            destroyStrategy = DEFAULT_DESTROY_STRATEGY;
-          }
+            // Either set in form definition, or as part of globalOptions.
+            var destroyStrategy =
+                !form.hasOwnProperty('destroyStrategy') ? DEFAULT_DESTROY_STRATEGY : form.destroyStrategy;
+            var schemaType = getSchemaType();
 
-          if (destroyStrategy === 'retain') {
-            return; // Valid option to avoid destroying data in the model.
-          }
-
-          destroyUsingStrategy(destroyStrategy);
-
-          function destroyUsingStrategy(strategy) {
-            var strategyIsDefined = (strategy === null || strategy === '' || strategy === undefined);
-            if (!strategyIsDefined){
-              strategy = DEFAULT_DESTROY_STRATEGY;
+            if (destroyStrategy && destroyStrategy !== 'retain') {
+              // Don't recognize the strategy, so give a warning.
+              console.warn('%s has defined unrecognized destroyStrategy: \'%s\'. Used default instead.',
+                  attrs.name, destroyStrategy);
+              destroyStrategy = DEFAULT_DESTROY_STRATEGY;
             }
-            sfUnselect(scope.form.key, scope.model, strategy);
-          }
+            else if (schemaType !== 'string' && destroyStrategy === '') {
+              // Only 'string' type fields can have an empty string value as a valid option.
+              console.warn('%s attempted to use empty string destroyStrategy on non-string form type. ' +
+                  'Used default instead.', attrs.name);
+              destroyStrategy = DEFAULT_DESTROY_STRATEGY;
+            }
 
-          function getSchemaType() {
-            var sType;
-            if (form.schema) {
-              sType = form.schema.type;
+            if (destroyStrategy === 'retain') {
+              return; // Valid option to avoid destroying data in the model.
             }
-            else {
-              sType = null;
+
+            destroyUsingStrategy(destroyStrategy);
+
+            function destroyUsingStrategy(strategy) {
+              var strategyIsDefined = (strategy === null || strategy === '' || strategy === undefined);
+              if (!strategyIsDefined) {
+                strategy = DEFAULT_DESTROY_STRATEGY;
+              }
+              sfUnselect(scope.form.key, scope.model, strategy);
             }
-            return sType;
+
+            function getSchemaType() {
+              var sType;
+              if (form.schema) {
+                sType = form.schema.type;
+              }
+              else {
+                sType = null;
+              }
+              return sType;
+            }
           }
         });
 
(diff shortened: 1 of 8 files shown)


src/directives/schema-validate.js
```diff
@@ -1,110 +1,173 @@
...
+          } else {
+            // Angular 1.2
+            ngModel.$setViewValue(ngModel.$viewValue);
+          }
+
+        });
+
+
+        var DEFAULT_DESTROY_STRATEGY;
+        if (scope.options && scope.options.formDefaults) {
```
davidlgj: `formDefaults` should not be checked directly, it's already copied to the form object, see https://github.com/Textalk/angular-schema-form/blob/development/src/services/schema-form.js#L70

But in this case with `null` as a valid option distinct from `undefined` it might be better to move this to a global option. i.e. instead of checking `scope.option.formDefaults.destroyStrategy` just check `scope.option.destroyStrategy`

src/directives/schema-validate.js
```diff
@@ -1,110 +1,173 @@
...
+
+          function destroyUsingStrategy(strategy) {
+            var strategyIsDefined = (strategy === null || strategy === '' || typeof strategy == undefined);
+            if (!strategyIsDefined){
+              strategy = DEFAULT_DESTROY_STRATEGY;
+            }
+            sfUnselect(scope.form.key, scope.model, strategy);
+          }
+
+          function getSchemaType() {
```
davidlgj: Shouldn't this return schemaType instead of just settting it? It looks like it's used that way anyway on row 128.

src/directives/schema-validate.js
```diff
@@ -1,110 +1,173 @@
...
+          }
+
+          if (destroyStrategy === 'retain') {
+            return; // Valid option to avoid destroying data in the model.
+          }
+
+          destroyUsingStrategy(destroyStrategy);
+
+          function destroyUsingStrategy(strategy) {
+            var strategyIsDefined = (strategy === null || strategy === '' || typeof strategy == undefined);
```
davidlgj: Uhmmm... shouldn't `typeof strategy == undefined`  be `typeof strategy === 'undefined'` or  `strategy === undefined` or `angular.isUndefined(strategy)

davidlgj:
Hi @jbsaff and thank you for the PR! This is a much sought after feature and even though it changes current behavior I think that is a good call. It's probably what you want anyway and its the current behavior that surprises people. Nice work :+1: 

I made some comments, I think it's a good idea to add a global option for it instead of checking formDefaults explicitly. It was unfortunate that the change in indentation broke the diff :) was a bit hard to read!

src/directives/schema-validate.js
jbsaff: Looking at this again, I think I might have overlooked the implications of having the formDefaults copied to the form object. (It makes sense now, so I may have just had the wrong idea when I started trying to use it). It seemed like a convenient way to set the default for the whole form if someone wanted to use an alternative to undefined, but a standard global option accomplishes that. I'll make a change.

src/directives/schema-validate.js
jbsaff: Yes. D'oh. Will fix.

src/directives/schema-validate.js
jbsaff: Will fix.

jbsaff:
Thanks for the feedback. I'm pretty comfortable in Java, but needed something client-side for my current project. I'll incorporate your feedback into another commit. Thanks for your work on ASF. It's going to save me a ton of time!

src/directives/schema-validate.js
davidlgj: Awesome! :)

jbsaff:
@davidlgj Fixes checked in, and I think I got my branch rebased to match Textalk/development. 

Both the code and documentation should be updated to include the globalOption and form defined destroyStrategy. I had to rework my approach to account for the difference between having the destroyStrategy property declared as undefined and not having it declared at all.

Thanks!

davidlgj:
:+1: Awesome! I'm hoping to test and merge start of next week :) 

One thing that struck me though, what happens if you remove the entire form? Like when navigating away in a single page app, or an ng-if on something around the form. Won't that trigger the `$destroy`?Will that then remove all the attributes from the model?

jbsaff:
Well crud. Yes, if there is an ng-if wrapping the form, the entire form would be subject to that $destroy.  I'll look for another solution.

jbsaff:
Okay, I've given it a thought. There are actually two parts to the logic that ought to trigger data cleanup. One is removal from the DOM, which can be identified by the $destroy. The other is that the condition on the form element is not satisfied (which might trigger a $destroy due to how conditions are wired in with ng-if). I'm going to head in that direction and see how it goes.

davidlgj:
I was thinking that there might be a solution if the $destroy is called on
the first on sf-schema $scope then we might be able to set a flag that
stops the destruction of the fields.

fre 1 maj 2015 15:22 James Saffell <notifications@github.com> skrev:

jbsaff:
Trying to understand the implications of the Angular docs for the $destroy event. It says that the listener is called just prior to the actual destruction occurring. Does that mean that it happens after all of  the children have finished processing the $destroy in a depth-first fashion? If so, the $on $destroy for sfSchema would be too late to set the flag, as all of the children have already gone. I'm going to try to improve my understanding of the situation and determine what the event firing order is.

jbsaff:
I've put together an initial (and possibly naive) implementation of a flag to retain the model instead of allowing the destroyStrategy to be used when the sfSchema element is destroyed.

davidlgj:
I'll merge it locally and test it out!

src/directives/schema-validate.js
```diff
@@ -1,110 +1,190 @@
...
+          return defaultStrategy;
         }
 
-      });
+        // Clean up the model when the corresponding form field is $destroy-ed.
+        // Default behavior can be supplied as a globalOption, and behavior can be overridden in the form definition.
+        scope.$on('$destroy', function() {
+
+          var form = getForm();
+          var conditionResult = $parse(form.condition);
```
davidlgj: I was looking through the code and testing some, and I'm not quite sure why we need to check on the condition here? Shouldn't we always just follow the destroy strategy if we get a destroy event? I'm thinking that condition might not be the only thing that can remove a form element from the form, any other add on might use a `ng-if` somewhere and since any field not in the DOM won't get any validation done on it anyways It's probably best to remove it from the model.

davidlgj:
Ok, so I didn't get through i tonight, I'll try to pick it up again soon. Keeping the flag in a service won't work since that service is shared across all instances of schema form you might have on the page, but I think I got a solution for that :) ...just got to try it out first!

src/directives/schema-validate.js
jbsaff: I had left it in as a way to help show where I was coming from with my solution, because I wasn't 100% confident that the service I put in would be correct (I'd still label myself an apprentice with Javascript). Assuming that the flag will handle for the case where the destroy is coming from outside of the form, then I agree that destroys initiated within the form should just follow the destroyStrategy. It'll be more consistent (and still configurable) and easier to explain.

jbsaff:
Okay, thanks for the assist. I appreciate the help polishing it up!

davidlgj:
@jbsaff  I've started to merge it in, but I've also changed some things. I really hope you don't mind. You've done some solid work and I couldn't have got this done without your PR to base it on! 

Initial manual testing looks good, I pushed the branch so you can see it here if you feel like checking it out: https://github.com/Textalk/angular-schema-form/tree/jbsaff-feature/cleanModelUsingDestroyStrategy

It needs some cleaning and some tests but I think it works :)  

Major points of what I have changed:
  * I changed the option `defaultStrategy` value to always take a string, so its "null", "empty" (was ''), "remove" (was undfined) and "retain". This greatly simplified the logic when trying to figure out if you should inherit a global option or not.  
  * I changed the use of a service to flag for "external destruction", instead I put it as a variable on scope. This means several instances of schema form won't clash with each other.

jbsaff:
@davidlgj I'm glad to hear my work was a useful starting point. I try not to take any ego into a project; if my stuff is good, it will stay. The changes you described sound like they will make the feature far more useable within ASF, so I'm very much in favor. Thanks for the reviews and suggestions. It made my first contribution to a project a lot of fun.

jbsaff:
Only thing I noticed is that when an array element (like in a tabarray) is removed, the console fills up with "TypeError: Cannot convert undefined or null to object". If I set the global option to null, I get "TypeError: Cannot set property '[object Array]' of undefined" instead. It looks like the Remove cleans up the model (as expected), then the new feature tries to do the same thing and gets an error. Do we expect that, in a case where sfSelect would return undefined, we would still want to attempt to set the value using the destroyStrategy?

davidlgj:
Of course! We might not have a model if the user hasn't filled out the entire form :) Added  a check! 
This seems to work for me, but in certain scenarios, like when having nested objects we can end up being left with empty object `{}`, since there is no field with a `schema-validate` directive at that level... I'm gonna try and move it to the decorator directive instead. 
 
btw I've been testing with this schema and form definition: https://gist.github.com/02ed18af7570ef149c92

davidlgj:
Ok, moved it around a bit and added some tests. I tried doing removal of empty objects, but that is a bit tricky. Could easily end up mucking up arrays by removing the empty object needed for it to display an empty form. I'm not convinced it's actually needed, so I'm gonna let it be as it is. I think its ready enough for merging into the development branch :) Thanks for the PR and all the help!

Anthropic:
Nice work @jbsaff & @davidlgj look forward to using it and replacing my own workarounds :)

jbsaff:
@davidlgj  Having the behavior in the decorator directive looks like it works for the case where the destroy is triggered by something outside the form. However, when a form field destroy is triggered by a condition no longer being satisfied within that form, the sf-decorator tag remains in the DOM and the cleanup of the model isn't triggered. 

Would it make sense to split the behavior off into a directive that can be applied to the outermost div in each decorator template? It would require add-ons/extensions to opt-in to using the new directive, which could be a drawback. I like the way that moving the event handling up from the schema-validate level means that it will only fire once per field, instead of once for every input (long checkbox/radio lists, commonly).

davidlgj:
@jbsaff You're absolutely right! A separate directive is probably the cleanest, I'll try to re-work soon.

Issue #295, opened 2015-03-12
onChange doesn't fire when form type is array


whoadave:
As the title says, if I add an onChange handler to an array field, and then add or remove an item to that array, the handler doesn't fire.

nicklasb:
Fixed, closing.

nicklasb:
This is not fixed, the pull request in #295 has not been merged.

Anthropic:
I believe this is fixed in the latest updates (alphas) for the eventual 1.0.0 release.

Please re-open if there are still issues in these versions.

Anthropic:
Hmmm the test case in the originally proposed PR fails, re-opening just in case it isn't fixed.

Anthropic:
Nope just fixed the test and it does pass. All good.

santhoslokanath:
Hi Team,

In which version of release this patch is available. I am using bower to install and after running bower install. the patch code mentioned in this case(i.e onChange for array) is not available.

Please, let me know what I have do to get patch through bower.

Thanks,
Santhosh Lokanath

Anthropic:
@santhoslokanath as mentioned above it is fixed in the alphas, it is not in a release yet, I suggest using npm and pulling the alpha version. Bower won't be supported much longer I expect as they recommend using yarn on their own homepage https://bower.io/

Issue #208, opened 2015-01-12
onClick event on array "+ add" button


Simomultimedia:
Is it possible to make `onClick` work on array buttons?

```
        {
          "key": "Questions",
          "add": "New Question",
          "onClick": "myFn()",
       }
```
```
$scope.myFn = function() {
   console.log('blah blah')
}
```

davidlgj:
I guess, but why? Whats the use case? It's a bit fiddly since it already has a ng-click that adds an instance to the array.

Simomultimedia:
I want to create a list that will be updated everytime "+add new" is clicked

davidlgj:
Updated in what way? From server? What about the instance being added? 

Do a workaround otherwise, hide the add button with css and put an button below with an onClick on it?

israelidanny:
Actually this is required for me too. My use case is a form that will be displayed via an iFrame. 
I need to hook myself onto anything that changes the document size and report it to the parent window.

airdrik:
I've also got a use for this (or similar):
I've got a schema for json documents which describe a grid-like structure with rows and columns.  Some of the rows only contain some of the columns.  
In the sub-form for the rows I want to display a list of the columns so the user can select which columns the row contains.  As such I need to keep the list of columns in the rows sub-form in sync with the list of columns from the columns sub-form.
e.g.
```javascript
    [
        { key: "columns", type: "tabarray",
            onChange: updateRowColumns() // or onAdd/onRemove, or whatever
        },
        { key: "rows", type: "tabarray", items: [
            "rows[].rowIdentifier",
            {
                key: "rows[].columns", // currently implemented as a list of indexes into the columns array
                type: "checkboxes",
                titleMap: buildFromColumns() // list of {value: columnIdx, name: column.name} for the checkboxes
            }
        ]}
    ]
```

airdrik:
Is there any way to listen to changes to the size of an array or tabarray form element? 
Or would it be easier to set up some kind of watch on the underlying array object?

davidlgj:
There is currently no way to listen to changes of the array, so a $watch on the underlying object is your only choice until onChange is implemented for arrays.

PR #374, opened 2015-04-29
#295 array on change

luciformed:
fix for #295, also #208 ? by calling ngModel.$setViewValue with a copy of array


davidlgj:
Hi @luciformed, sorry for the late reply and thanks for the PR!

I'm a bit worried about the `angular.copy`, especially if the array items part of the form is large and complex. I'm gonna try it out :-)

luciformed:
angular documentation is where I read you should pass copies to `$setViewValue`, i understand your concern with performance though.

luciformed:
dod you get the chance to give it a try?

davidlgj:
Sorry it's taking so long. I hope to get time to dig into this on friday.

davidlgj:
I see that you set the view value to trigger a the `ngModel.$viewChangeListeners` that `sf-changed` use.  

I'm thinking it could be simpler to just call the onChange listener instead, that is. Something like the code here: https://github.com/Textalk/angular-schema-form/blob/development/src/directives/changed.js#L19

By the way, would you mind removing the `dist/` folder and its files from PR?

oatkiller:
Any progress with this? I need type: array and I'm adding custom validation errors representing backend errors. This is killing me :(

grjones:
I would love for this to be merged in.  It fixed our problems. :+1:

luciformed:
sorry I forgot this was still hanging, I'll update the PR and hopefully it gets merged

Anthropic:
@luciformed thanks for the PR, I've started maintaining the repo and starting to try and get on top of neglected PRs, could you do me a favour and **remove the dist folder** from the PR please?

mmgm:
I was just about to submit an almost identical pull request (frankly with less tests, so this one is better) that solved our problem.

From my understanding this needs to also come with a pull request to angular-schema-form-bootstrap that removes sf-changed from https://github.com/json-schema-form/angular-schema-form-bootstrap/blob/develop/src/checkboxes.html

Is this correct? Is there anything I can do to help this get merged? @Anthropic

Anthropic:
@mmgm @luciformed needs to remove the dist folder from the PR, ideally the changes would now be done against the webpack/babel branch too as that is soon to be merged into the main develop branch. But I need to write some documentation on how to work with that before then.

If you could pull @luciformed 's version to your own fork and then make a PR minus dist folder I could accept that if this one isn't updated.

nicklasb:
@mmgm @luciformed, have you considered the proposal from @Anthropic?

Anthropic:
I added your test case @luciformed and it now passes in the new version, thanks so much for making the test case :)

PR #387, opened 2015-05-12
Simplify sfMessage directive and uses

mike-marcacci:
There's no need to overthink sfMessage; in all cases it uses the parent scope's `form.description`. This allows faster processing, and full data-binding so that an application can modify descriptions on-the-fly.

### DIFF
--- a/src/directives/message.js
+++ b/src/directives/message.js
@@ -1,53 +1,28 @@
 angular.module('schemaForm').directive('sfMessage',
-['$injector', 'sfErrorMessage', function($injector, sfErrorMessage) {
+['sfErrorMessage', function(sfErrorMessage) {
   return {
     scope: false,
     restrict: 'EA',
-    link: function(scope, element, attrs) {
+    template: '<div ng-if="!error && form.description" ng-bind-html="form.description"></div><div ng-if="error" ng-bind-html="error"></div>',
+    link: function(scope) {
 
-      //Inject sanitizer if it exists
-      var $sanitize = $injector.has('$sanitize') ?
-                      $injector.get('$sanitize') : function(html) { return html; };
+      scope.$watchCollection('ngModel.$error', function(errors) {
 
-      //Prepare and sanitize message, i.e. description in most cases.
-      var msg = '';
-      if (attrs.sfMessage) {
-        msg = scope.$eval(attrs.sfMessage) || '';
-        msg = $sanitize(msg);
-      }
+        errors = Object.keys(errors);
 
-      var update = function(valid) {
-        if (valid && !scope.hasError()) {
-          element.html(msg);
-        } else {
+        // We only show one error.
+        // TODO: Make that optional
+        var error = errors[0];
 
-          var errors = Object.keys(
-            (scope.ngModel && scope.ngModel.$error) || {}
-          );
-
-          // We only show one error.
-          // TODO: Make that optional
-          var error = errors[0];
-
-          if (error) {
-            element.html(sfErrorMessage.interpolate(
-              error,
-              scope.ngModel.$modelValue,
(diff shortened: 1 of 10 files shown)


src/directives/message.js
```diff
@@ -1,55 +1,30 @@
...
-        }
-      };
-      update();
-
-      scope.$watchCollection('ngModel.$error', function() {
-        if (scope.ngModel) {
-          update(scope.ngModel.$valid);
-        }
+    template: '<div ng-if="!error && form.description" ng-bind-html="form.description"></div><div ng-if="error" ng-bind-html="error"></div>',
+    controller: ['$scope', function($scope) {
```
davidlgj: Just curious, why controller instead of link?

src/directives/message.js
```diff
@@ -1,55 +1,30 @@
...
-          }
-        }
-      };
-      update();
-
-      scope.$watchCollection('ngModel.$error', function() {
-        if (scope.ngModel) {
-          update(scope.ngModel.$valid);
-        }
+    template: '<div ng-if="!error && form.description" ng-bind-html="form.description"></div><div ng-if="error" ng-bind-html="error"></div>',
```
davidlgj: I see the simplification here, but I'm not sure if I like it having a template, although I'm not very fond of direct placement of description with `element.html()` either, also feels like a hack. 

The thing is that a another decorator might want to have another html here, or display the description in another place.

davidlgj:
Hi @mike-marcacci, thanks for the PR! 

One thing though, would you mind redo it without the files in `dist/`? It's a bit of a hassle to merge and it mucks up the history. 

I've commented a bit above as well, am I right to presume its the two way binding with description that is the main problem here?

mike-marcacci:
Hey @davidlgj sorry for the slow reply! Been a crazy week. Yes, you're correct that the auto-binding is what I'm really after. I get what you're saying about custom add-ons wanting to use a field with a name other than `description`. However, I might contest that non-default behavior should be a little more difficult to implement than the default behavior (it's nice having description standardized), and thanks to the fact that the `sfErrorMessage` service does all the heavy lifting, it would be almost trivial to copy and modify the `sfMessage` directive's template and controller.

I'll go ahead and remove the `dist/` files in case this solution ends up working out!

src/directives/message.js
mike-marcacci: It would actually be better written as link – I'll go ahead and do that. My first iteration actually used both (one to compile the element, the other the way it is). I'm going to change this back to link :) good catch!

davidlgj:
I've started some tentative work on an angular material decorator, I'm going to see where that goes, but it might be that we don't want descriptions at all there! Anyways, if I don't merge this as it is I'll at least try to fix up the binding.

davidlgj:
Hi @mike-marcacci 

Yes, so currently in the budding material decorator, https://github.com/Textalk/angular-schema-form-material we're using sf-message *without* description at all. Would it be OK to reject this PR if I implement a watch on the description so its always updated?

mike-marcacci:
Absolutely! Thank so much for the politeness :) I'm really excited to see how that material decorator, ahem, materializes

davidlgj:
Great! I'll try to get the watch done today. Checkout the development branch, it's starting to come together, inputs, textarea, selects and checkboxes sort of work. @Anthropic is doing some awesome work on it!

Dervisevic:
> I'm really excited to see how that material decorator, ahem, materializes

Puntastic

davidlgj:
@mike-marcacci 

I've made a watch on description in development branch, https://github.com/Textalk/angular-schema-form/commit/e7ff94b4b05cd9f4042b6c51fe4b68edffa56ce7

Seems to work(TM) :)

PR #394, opened 2015-05-18
Add ability to use named enums

megawebmaster:
Hi,

I found Schema Form was lacking ability to set nice titles for enums in schema definition. This was somewhat hard for me, as I was unable to redefine a key using form definition (like it's shown in docs) as this solution was adding another element to rendered form (I was using an asterisk to render all elements). Here's my proposition:

Schema:
```
{
    "type":"object",
    "properties":{
        "direction":{
            "type":"string",
            "title":"Kierunek relacji",
            "enum":{ "directed":"Directed","bidirectional":"Bidirectional","none":"No direction"}
        }
    }
}
```

This renders as it should and I have abilities to set nice names.


Dervisevic:
Doesn't [x-schema-form](https://github.com/Textalk/angular-schema-form/blob/development/docs/index.md#form-defaults-in-schema) work for this with [titleMap](https://github.com/Textalk/angular-schema-form/blob/development/docs/index.md#select-and-checkboxes)? Or are you missing some feature?

megawebmaster:
Well, it seems I hadn't understand this section: works perfectly, thank you ;)

Dervisevic:
There's so much that it isn't always easy to see what you can do with it :)

Dervisevic:
Btw join us over at https://gitter.im/Textalk/angular-schema-form if you have any questions or thoughts.

megawebmaster:
OK, no more questions for now - but they will probably emerge as I'm using this library extensively.

PR #396, opened 2015-05-19
Global option propagation

golfmat1:
Small change to get global options to propagate through form


davidlgj:
Hi @golfmat1 and thanks for the PR! Small but good :+1:

Issue #203, opened 2015-01-08
Ability to hide title


stramel:
It would be great to be able to hide the `title`. Having no title on schema and form should result in no `fieldset` `title`. 

```js
// Schema
name: {
  type: 'object',
  properties: {
    first: {
      title: 'First Name',
      type: 'string'
    },
    last: {
      title: 'Last Name',
      type: 'string'
    }
  }
}
```

```js
// Form
{
  key: 'name',
  items: [
    {
      title: 'FIRST_NAME',
      key: 'name.first',
      type: 'text',
      placeholder: 'FIRST_NAME',
      validationMessage: {}
    }, {
      title: 'LAST_NAME',
      key: 'name.last',
      type: 'text',
      placeholder: 'LAST_NAME',
      validationMessage: {}
    }
  ]
}
```

Anthropic:
Does, {"notitle": true} in the [Standard Options](https://github.com/Textalk/angular-schema-form/blob/master/docs/index.md#standard-options) not do this? How should it be different?

davidlgj:
Maybe "supressPropertyTitles" in [Global Options](https://github.com/Textalk/angular-schema-form/blob/master/docs/index.md#global-options) would work for you?

stramel:
@Anthropic @davidlgj I will check this today, thank you for the suggestions! Also, I think having the styling options available in the Standard Options will help as well.

stramel:
Fieldsets do not seem to be respecting the `notitle` property. 
https://github.com/Textalk/angular-schema-form/blob/master/src/directives/decorators/bootstrap/fieldset.html#L2

ricklxm:
"Fieldsets do not seem to be respecting the notitle property. ". I found this is true? Is there any solution to workaround this?

davidlgj:
@ricklxm  Basically its just a `ng-hide="form.notitle"` that is needed on the legend tag. It just needs to be done :)

stramel:
@ricklxm @davidlgj Please take a look at my PR #414 if you get a chance. I was just too lazy to do it when I originally posted lol

davidlgj:
Thanks @stramel! Your PR is now in the development branch.

PR #414, opened 2015-06-03
fix(fieldset): make fieldsets properly respect notitle

stramel:
This fix should visibly hide the title on fieldsets while allowing screen readers to read them if they exist.

NOTE: I can change this to a simple `ng-show="showTitle()"` instead if that would be better.

Addresses #203


davidlgj:
:+1:  Thanks!

PR #418, opened 2015-06-04
Add parameter `trackBy` on form type select.

Fridus:
With this feature, it's now possible to set like value an object in `titleMap` and add its property what we want too.
`trackBy` parameter used to specify the fields on which we want to compare this the model.

Ex:
```
scope.thing = { id: 2, a: 1, b: 2 };

scope.schema = {
  "type": "object",
  "properties": {
    "thing": {
      "title": "Thing"
    }
  }
};

scope.form = [{
  key: "thing",
  type: 'select',
  titleMap: [{
    id: 1, name: 'abc', value: { id: 1 }
  },{
    id: 2, name: 'def', value: { id: 2 }
  },{
    id: 3, name: 'ghi', value: { id: 3 }
  }],
  trackBy: 'id'
}];
```

### DIFF
--- a/test/directives/schema-form-test.js
+++ b/test/directives/schema-form-test.js
@@ -1649,6 +1649,85 @@ describe('directive',function(){
   });
 
 
+  it('should show the correct option selected', function() {
+    inject(function($compile,$rootScope){
+      var scope = $rootScope.$new();
+      scope.thing = { id: 2, a: 1};
+
+      scope.schema = {
+        "type": "object",
+        "properties": {
+          "thing": {
+            "title": "Thing"
+          }
+        }
+      };
+
+      scope.form = [{
+        key: "thing",
+        type: 'select',
+        titleMap: [{
+          name: 'abc', value: { id: 1, a: 2 }
+        },{
+          name: 'def', value: { id: 2, a: 1 }
+        },{
+          name: 'ghi', value: { id: 3, a: 3 }
+        }]
+      }];
+
+      var tmpl = angular.element('<form sf-schema="schema" sf-form="form" sf-model="thing"></form>');
+
+      $compile(tmpl)(scope);
+      $rootScope.$apply();
+
+      setTimeout(function() {
+        tmpl.children().eq(0).find('select').eq(0).val().should.be.eq(2);
+      }, 0);
+
+    });
+  });
+
+
+  it('should show the correct option selected with `track by`', function() {
+    inject(function($compile,$rootScope){
+      var scope = $rootScope.$new();
+      scope.thing = { id: 2, a: 1, b: 2 };
+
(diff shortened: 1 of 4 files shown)


davidlgj:
Hi @Fridus, nice PR! Thank you for that :+1: Always nice to see tests as well :)

Fridus:
@davidlgj :) When will you make a new release ?

davidlgj:
Hmmm... the plan was to get one out by mid august. The thing is that I have begun working on a new "builder". So it's a big rewrite, but it will make it a lot better. But on the same hand it's been a long time since last release. 

So I actually don't what to do. The new "builder" stuff in development branch isn't finished yet, but its supposed to be backwards compatible so in theory  it *should* work, but I don't have enough test to be confident it does. Maybe I could try to make a beta release soonish and if that works make a proper release...

Fridus:
I ask you this because I need this feature fairly rapidly for a project :)

davidlgj:
@Fridus just made a release! hope it works for you.

davidlgj:
Hi @Fridus, I had to revert you commit. It broke the select when not using `trackBy`, see #462, #471, #473 

It looks like using `select as` in combo with `trackBy` is not supported. Also I now noticed that two of the tests lacked a `done` callback, so they seem to work but in fact fail. I didn't have time to fix it so I did a revert.

test/directives/schema-form-test.js
```diff
@@ -1649,6 +1649,85 @@ describe('directive',function(){
...
+        }]
+      }];
+
+      var tmpl = angular.element('<form sf-schema="schema" sf-form="form" sf-model="thing"></form>');
+
+      $compile(tmpl)(scope);
+      $rootScope.$apply();
+
+      setTimeout(function() {
+        tmpl.children().eq(0).find('select').eq(0).val().should.be.eq(2);
```
davidlgj: This need a `done` callback

test/directives/schema-form-test.js
```diff
@@ -1649,6 +1649,85 @@ describe('directive',function(){
...
+        trackBy: 'id'
+      }];
+
+      var tmpl = angular.element('<form sf-schema="schema" sf-form="form" sf-model="thing"></form>');
+
+      $compile(tmpl)(scope);
+      $rootScope.$apply();
+
+      setTimeout(function() {
+        tmpl.children().eq(0).find('select').eq(0).val().should.be.eq(2);
```
davidlgj: This one as well

Fridus:
Hi @davidlgj , I'll work on it when I have a little more time ;)

davidlgj:
:+1:

mån 13 jul 2015 10:55 Florent <notifications@github.com> skrev:

PR #420, opened 2015-06-04
Update list variable when watching for array changes

lukebennett:
In 5c6630b1899af38f1a302454ef87e40b600a0888, watching of arrays was refactored to remove a call to `sfSelect`, however in the process the `list` variable stopped getting updated. This led to adverse behaviour when adding new array items in the editor due to `list` and `scope.modelArray` being out of sync. This PR ensures the `list` variable is kept in sync with `scope.modelArray` when the array changes.


davidlgj:
Oh, I didn't realize I broke something! Thank you @lukebennett

PR #423, opened 2015-06-05
Fix custom validations on 1.2

eknowles:
This was breaking in Angular 1.2 because the browser dies on $setDirty as this method can not be called as it doesn't exist.

The fix is to check if the version is <= 1.2 and perform the .$dirty / .$pristine on the model.

### DIFF
--- a/dist/schema-form.js
+++ b/dist/schema-form.js
@@ -512,12 +512,11 @@ angular.module('schemaForm').provider('schemaFormDecorators',
                       }
 
                       if (scope.ngModel && error) {
-                        if (scope.ngModel.$setDirty()) {
-                          scope.ngModel.$setDirty();
-                        } else {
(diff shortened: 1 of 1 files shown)


davidlgj:
Hi @eknowles  and thanks for the PR!

But I can't merge this, the files in `dist/` are built from the sources, so any changes to it will be overwritten. 

But I did fix the bug, you don't need to check version, the `if (scope.ngModel.$setDirty())` was there to check for the presence of `$setDirty` but the extra `()` meant that it got called instead, so now it's changed to `if (scope.ngModel.$setDirty)`. see: 
  
https://github.com/Textalk/angular-schema-form/blob/development/src/services/decorators.js#L231

eknowles:
Perfect! Thank you!

PR #431, opened 2015-06-15
Fixing schema.{min,max}Items message mixup.

leipert:
Currently maxItems is shown for to short arrays and minItems for to long arrays.
It should be exactly the other way around:

screen shot 2015-06-15 at 18 24 59


Dervisevic:
Woops

Woops! Good catch, and thanks for the PR!

leipert:
No problem, they call me the code review :grin: 
Thanks for the fast merge.

PR #441, opened 2015-06-22
Revalid before apply tv4 validation, fix #440

### LINKED ISSUE #440: Problem: when use $scope.$broadcast('schemaForm.error.key', errorCode, true) the field still marked invalid.
here's my usecase:

2 date field, one for begin date, one for end date, the custom validator try to notify user the beginDate should less than endDate, and both of them are required fields.

```
$validators: {
              beginLessThanEnd: function (v) {
                // following just sample code, not real working code
                var valid = beginDate < endDate
                $scope.$broadcast('schemaForm.error.' + (key === 'endDateTime' ? 'beginDateTime' : 'endDateTime'), 'beginLessThanEnd', valid);
                return valid;
              }
}
```

When I set beginDate larger than endDate, it does correctly mark 2 fields with the error message I setup, when I set beginDate back to smaller date, it get both fields valid, it's good. but, if instead, I set the endDate to a larger date then the beginDate, the endDate will be valid, but the beginDate is still invalid.

The reason of these is that the beginDate field got the "undefined" value at the initial validation failure, when the endDate get larger, the beginDate is still have "undefined" modelValue (but the viewValue is what you see in browser), then it use the undefined value to check validation (inited by the "scope.$broadcast('schemaFormValidate');" call inside the schemaForm.error broadcast handler), it will result an 302 required error of tv4.

Any ideas to resolve this? I'm using the current master branch code. with angular 1.3.14

ulion:
before we re-valid the field value, we need call $validate() to re-push the viewValue to the validators, else we will always use the undefined modelValue if previous validation is failed due to any of the validators but currently it can pass it.

### DIFF
--- a/src/services/decorators.js
+++ b/src/services/decorators.js
@@ -247,7 +247,10 @@ angular.module('schemaForm').provider('schemaFormDecorators',
 
                         scope.ngModel.$setValidity(error, validity === true);
 
-                        if (validity === true) {
+                        if (validity === true && !scope.ngModel.$valid) {
+                          // Previous the model maybe invalid, then the model value is undefined.
+                          // To make the validation to the tv4 works, we need also re-trigger validators
+                          scope.ngModel.$validate();
                           // Setting or removing a validity can change the field to believe its valid
                           // but its not. So lets trigger its validation as well.
                           scope.$broadcast('schemaFormValidate');


ulion:
it seems caused some deadloop, for certain conditions, any idea to resolve this?

ulion:
finally use form.ngModelOptions.allowInvalid to overcome this.

Issue #443, opened 2015-06-24
custom styling of validation messages is not available


prabhatsharma:
Validation mesages were appearing a bit crazily in our application due to various styles and I realized that help-block class is hard coded for styling of validation messages. We need support for custom styling of validation messages.

I have forked the repository and provided a fix for this. Will be raising a pull request.

prabhatsharma:
Have added following pull request
https://github.com/Textalk/angular-schema-form/pull/444

nicklasb:
Closing, the discussion may continue at the PR.

PR #444, opened 2015-06-24
Validation message styling through validationHtmlClass

prabhatsharma:
Needed support for custom styling of validation messages . https://github.com/Textalk/angular-schema-form/issues/443


davidlgj:
Hi @prabhatsharma and thank you for the PR! 

I'm a bit reluctant adding another watch,could this be solved by CSS in combo with `htmlClass`?

prabhatsharma:
Understand your concern around adding something extra, but we have tried fixing the styling through various means (css, htmlClass and overriding help-block) with success in many cases. However all scenarios are not getting covered with these and overriding help-block impacts other fields as well and is not really a good idea.

Would appreciate your consideration in getting this incorporated. This would make styling a lot simpler for validation messages.

Anthropic:
@prabhatsharma I know it has been a long time, but with David no longer on the development can you please give me an idea of the use cases you feel this supports that couldn't be worked out in any other way, like perhaps in a plunker?

prabhatsharma:
Hi there, Idea of having this  PR was to be able to add styling to the validation message that happens on schemaform fields. In absence of this implementing custom styling is difficult and we had to resort to creating custom schemaform components. Schemaform already provides styling for label and field for teh component, but not for validation message.

nicklasb:
So I basically have one thing here and that is for pure traceability, I do not like that the code block has moved the way it have, it means that diffs involving this code will always be problematic because of this commit.
 
I couldn't/wouldn't even find the change. 
That is, with the very limited amount of effort I was prepared to put into it. :-)

Anthropic:
@prabhatsharma Nicklas is Referring to the dist folder, that will need to be reverted before any PR could be accepted.

When I asked about what you can't do, I mean if you use the parent tag it gets has-error and also names field type ie. .schema-form-section, so I am just wondering what scenario you are having trouble supporting with CSS?

Whatever you add to form.htmlClass you could use as:
```css
.my-css-class.has-error  .help-block {
}
.has-error .help-block {
}
```
If there are any specific templates that don't behave in a consistent way which made that difficult I would gladly accept a PR that made them behave in a way that made a simple CSS rule apply across all templates.

nicklasb:
@prabhatsharma, @Anthropic  I am sorry, and thanks for the save there @Anthropic but that was just me being an idiot. :-)
I didn't see it was the dist file, I thought it was the source file. Disregard my earlier statement.

(however it is kind of why we don't want dist files to be included in commits)

PR #447, opened 2015-06-25
Fix build on case sensitive systems

LoicMahieu:
In browserify context, when requiring `angular-schema-form` in a case sensitive system (my case ubuntu):
```
Browserify Error: Cannot find module 'ObjectPath' from '/home/gitlab-runner/tmp/builds/project-51/backend/node_modules/angular-schema-form/dist'
```

Obviously all dependencies name are always in lower case. (bower and npm)

### DIFF
--- a/gulp/tasks/minify.js
+++ b/gulp/tasks/minify.js
@@ -16,7 +16,7 @@ gulp.task('minify', function() {
     dependencies: function() {
       return [
         {name: 'angular'},
-        {name: 'ObjectPath'},
+        {name: 'objectpath'},
         {name: 'tv4'},
       ]
     },
--- a/src/sfPath.js
+++ b/src/sfPath.js
@@ -1,6 +1,6 @@
 angular.module('schemaForm').provider('sfPath',
 [function() {
-  var sfPath = {parse: ObjectPath.parse};
+  var sfPath = {parse: objectpath.parse};
 
   // if we're on Angular 1.2.x, we need to continue using dot notation
   if (angular.version.major === 1 && angular.version.minor < 3) {
@@ -8,7 +8,7 @@ angular.module('schemaForm').provider('sfPath',
       return Array.isArray(arr) ? arr.join('.') : arr.toString();
     };
   } else {
-    sfPath.stringify = ObjectPath.stringify;
+    sfPath.stringify = objectpath.stringify;
   }
 
   // We want this to use whichever stringify method is defined above,


thardyman:
It would be really good to get this fixed. I has caused me a lot of headaches, as it was working on my machine, but failing on our build server.  Both machines were running the same ubuntu image, the only difference was my local machine had a drive synced with my host OS - Mac OS.

thardyman:
@LoicMahieu did this PR fix your build issue?  I'm still getting problems using ASF with browserify on ubuntu.

davidlgj:
Hi @LoicMahieu and thanks for the PR!  I don't use browserify, how can I set up a test case that tests this? Why did you change the usage of `ObjectPath` name in `sfPath.js`, as you see from the tests it breaks, I guess it works with browserify?

thardyman:
Hi, sorry to hijack this thread, but the steps to reproduce are: -
```
> npm init
[accept all the defaults]
> npm install --save angular-schema-form
> echo "require ('angular-schema-form')" > index.js
> npm install browserify
> ./node_modules/browserify/bin/cmd.js -e index.js -o output.js

Error: Cannot find module 'ObjectPath' from '/home/timh/test/node_modules/angular-schema-form/dist'
```

I think the solution is to modify the minify.js gulp task, line 19 to: -

```
{name: 'ObjectPath', cjs: 'objectpath'},
```

davidlgj:
@thardyman no problem! Thanks

davidlgj:
@thardyman Thank you for the help! ok, so it seems you're right. Changing to `{name: 'ObjectPath', cjs: 'objectpath'},` fixes the build problem, but how do I test that the build actually works? i.e. what is the minimal index.html page with a browserify build?

thardyman:
Hmm, that's a good question! In theory browserify should have brought all the dependencies in to that `output.js` file we create in my _steps to reproduce_ example above.  So all we need is to add that via a script tag to a html page and we should be away.

Then it is just a case of what is the minimum markup / code for an ASF.

davidlgj:
I pushed the fix to development, now at least.

davidlgj:
Ok, I'll give it a try :)

davidlgj:
Ok. I think I got it now. When building with browserify its `objectpath` thats available to us, but `ObjectPath` when not using browserify.

thardyman:
@davidlgj thanks for figuring this out and getting it into a release so quickly!

PR #448, opened 2015-06-29
Fixing enum error Message

leipert:
If there is an error, the enum Property will actually not be set, so we need to use `{{viewValue}}` instead of `{{value}}`:

screen shot 2015-06-29 at 15 30 57


davidlgj:
Hi @leipert and thank you for the PR! Nice catch.

PR #463, opened 2015-07-07
Use fully qualified, unique values for FORs, IDs, and NAMEs.

gwbarrett:
Anyplace where there is a DOM field ID or NAME, or a label element FOR attribute, change it to use a fully qualified value. For IDs and FORs, join the key components with a '-', and for NAMEs, join key components with a '.'.

Previously, only the last component of the form.key was used to build these values (in most cases); the problem with that is when you have multiple leaves in your model with the same name for the final component.  In these cases, if you click on the label for any of these fields, focus is given to the first one instead of the current one as was probably intended.

Also, IDs are intended to be unique within the page DOM.


davidlgj:
Hi @gwbarrett, just so you know we haven't missed this PR, it's sort of the same as #496 and as soon as the new builder is up and running we'll make a solution based on both your PR's. We wan't  short version for `name` attribute though since using a FQN there makes life harder for browser autofillers. But that could be solved with an option :)

gwbarrett:
This issue has been around now for a long time.  Any progress?  This is a killer for me.

Anthropic:
@gwbarrett I'm planning on sorting out this issue for the next update, but it will have to target the bootstrap decorator. I have made a webpack-babel branch in that repo and will attempt to get your changes in there soon.

Anthropic:
The behaviour has been completely overhauled in the latest development branch code, the keys now have the full path hyphenated, so I will close this PR now. The only thing not covered is that if a form field is added multiple times the id's will be the same still. I am working on releasing the bootstrap update soon.

PR #469, opened 2015-07-08
Fix documentation typo

zerofifteen:
$validators example uses validationMessages (plural), when it should be using validationMessage (singular)


davidlgj:
Oh, my bad. Thank you @zerofifteen!

PR #474, opened 2015-07-10
Update title in array decorator template

prettymuchbryce:
This commit aims to make the array decorator template more idiomatic by replacing the `<h3>` element in the title, with the `<label>` element common to the other decorate template titles.

In addition this commit also replaces the inline `ng-show` logic with the `showTitle()` method, which looks to be equivalent.

This commit doesn't seem to impact any tests. I have ran them all and they pass.


davidlgj:
:+1: Thanks @prettymuchbryce

Issue #473, opened 2015-07-09
select doesn't show its initial value


aromanovich:
Here is a plunker to reproduce the bug: http://plnkr.co/edit/hGJJAZpd67uVpJlJBYlt?p=preview

It seems to be introduced in 0.8.3 with the [b7fb23a](https://github.com/Textalk/angular-schema-form/pull/418/files#diff-b7fb23ab1fda485328f4ece15a745a38R12) commit.
Angular docs says `as` and `track by` must not be used together: https://docs.angularjs.org/api/ng/directive/ngOptions#-select-as-

nicklasb:
I agree, this is strange, there is an existing issue for this issue, enum is not initiated when the type in the form is explicitly set:
#462

AlexAlbala:
Yes, the error appeared to me last week, just when they introduced the "trackBy" element

nicklasb:
Sorry, removed the link above, this is a somewhat different bug, even though the symptoms are similar.

AlexAlbala:
@aromanovich is right, I just removed the "track by item[form.trackBy]" element from the template of the select and voilà, worked lilke a charm!

davidlgj:
This was due to https://github.com/Textalk/angular-schema-form/pull/418 which is reverted in release 0.8.4.

Issue #458, opened 2015-07-03
Selected option from enum does not update UI


hola-senior:
I am attempting to run the custom-validators example after modify it to include an enum:

```javascript
$scope.schema = {
     "type": "object",
      "properties": {
           "gender":  {
               "type": "string",
               "enum": ["male","female","other"]
           },
      },
};
$scope.form = ["*"];
```
The resulting page does not update after selecting a gender, but the underlying model does update correctly. I am using Firefox 35. Am I missing something?

waleedasif322:
Can you replicate it in a plunker or jsfiddle?

hola-senior:
I think I've narrowed down how to reproduce. The issue does not appear when using angular 1.4, but does when using 1.2: http://jsfiddle.net/y4hv6w27/2/

agzam:
I can confirm, using ng1.3.16 here. And it's not working.

AlexAlbala:
I am using angular 1.3.14 and this problem has appeared last week for me...

mike-marcacci:
Hey guys! I'm hitting this too, and it's blocking something I need to do today/tomorrow so I'll go ahead and fix it... unless somebody else is already working on it (in which case I'd love to work off of your branch/fork). Anybody have a bit more insight on this?

agzam:
Any updates on the issue guys?

davidlgj:
@mike-marcacci @agzam @AlexAlbala @hola-senior @waleedasif322 
It seems like it was the `trackBy` feature that broke this, PR https://github.com/Textalk/angular-schema-form/pull/418

 I just reverted it and released 0.8.4

mike-marcacci:
Wonderful! Thanks for that @davidlgj – I was going to sit down to figure this one out right now :)

PR #475, opened 2015-07-12
Revert "Add parameter `trackBy` on form type select."

davidlgj:
Reverts Textalk/angular-schema-form#418

It caused a regression bug: https://github.com/Textalk/angular-schema-form/issues/473
Some of the tests also lacked a `done()` callback so they just appeared to succeed.


davidlgj:
It was also the culprit behind https://github.com/Textalk/angular-schema-form/issues/458

PR #486, opened 2015-07-27
minLength validation bug fix

iamdarrenhall:
Fix bug where minLength is incorrectly validated against maxLength


davidlgj:
Thank you @iamdarrenhall for spotting that one!

PR #493, opened 2015-07-29
Add npm/jspm-compliant index.js

nicklasb:
This commit adds module loader support for SystemJS/JSPM by defining entry points via require statements in an index.js-file.
The "filename" attribute in package.json seems like a non-standard package.json-attribute, anyone knows why?

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,4 @@
+/* */ 
+
+require("./dist/schema-form.min");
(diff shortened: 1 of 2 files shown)


nicklasb:
This is redundant, #495 should fix this as well.

nicklasb:
Perhaps this was reduntant after all. Again.

Issue #435, opened 2015-06-18
RequireJs wrapper for bootstrap-decorator


artemshapovalov:
Please add requirejs wrapper for bootstrap decorator. Because after concat vendor to one file Bootstrap decorator throws exception as:

```
Uncaught Error: [$injector:nomod] Module 'schemaForm' is not available! You either misspelled the module name or forgot to load it. If registering a module ensure that you specify the dependencies as the second argument.
http://errors.angularjs.org/1.3.15/$injector/nomod?p0=schemaForm
```

angelaharalson:
I'm having the same issue. Any work arounds?

robinduckett:
Really odd. Using browserify to include angular modules, schemaForm was "working" but throwing an error about not finding any decorators. This fixed it for me:
``` javascript
var schemaForm = require('angular-schema-form');
var bootstrapDecorator = require('angular-schema-form/dist/bootstrap-decorator');
```

Interestingly enough @angelaharalson I tried using your fork/commit, and nothing changed, and the above code worked.. Perhaps I'm requiring modules funny?

nicklasb:
Is this still a problem?

gornostal:
I'm seeing this problem with angular-schema-form#0.8.13

nicklasb:
Just a thought, are you using the external bootstrap decorator?
I had strange import problems that sort of magically went away when I switched to that.

(also, that is the way to use the framework, the internal one will permanently go away in 1.0)

gornostal:
Yes, I do.
First, I was confused by implicit dependency on `objectpath` and `tv4`, so I had to install them and include in a build process.
My gulpfile.js includes these files in a final build. `angular-schema-form-bootstrap/bootstrap-decorator.min.js` is among those files.
Do I do anything wrong?

```js
['bower_components/angular/angular.min.js',
 'bower_components/angular-route/angular-route.min.js',
 'bower_components/angular-ui-grid/ui-grid.js',
 'node_modules/angular-ui-bootstrap/dist/ui-bootstrap.js',
 'node_modules/angular-ui-bootstrap/dist/ui-bootstrap-tpls.js',
 'bower_components/angular-schema-form-bootstrap/bootstrap-decorator.min.js',
 'bower_components/angular-schema-form/dist/schema-form.min.js',
 'bower_components/objectpath/lib/ObjectPath.js',
 'bower_components/tv4/tv4.js']
```

nicklasb:
[...]

PR #495, opened 2015-07-30
Added UMD wrapper for bootstrap-decorator

angelaharalson:
Fixes this issue:
 RequireJs wrapper for bootstrap-decorator #435


nicklasb:
Interesting, I just made this PR:
https://github.com/Textalk/angular-schema-form/pull/493

:-)

nicklasb:
I realized that this is way better than my PR, as it should require() everything properly.

nicklasb:
While going on with packaging and modules, I should mention that I added ASF and some of the most used(?) ASF add-ons to the jspm registry:
https://github.com/jspm/registry/pull/510

nicklasb:
#493 turned out to not be redundant after all. :-(

davidlgj:
@angelaharalson awesome! I don't use RequireJS so I really appreciate help with this. Can I trouble you to redo the PR without the minified files in `dist`?

davidlgj:
@angelaharalson I merged it manually since I want to migrate the decorator to it's own repo. Thanks!

nicklasb:
Hi, this seems to be a problem with PR, I am however not sure what is the correct solution.
The problem is, that if one is using a module loading framework like SystemJS/JSPM or RequireJS, a dependency is on the "angular-schema-form"-form, as it is typically referring to a npm/bower/jspm registration, as you usually don't deviate from the naming, simply because one doesn't have to. 

This PR sets it to "schemaForm" for all situations, which does not work with those kinds of frameworks.

To make the component I am maintaining, [ngular-schema-form-dynamic-select](https://github.com/OptimalBPM/angular-schema-form-dynamic-select) work with my "MBE-frontend" which uses [SystemJS/JSPM](https://github.com/OptimalBPM/mbe-nodes/blob/master/example/static/jspm_config.js) and the names are automagically generated, I set the [amd and cjs options for the require](https://github.com/OptimalBPM/angular-schema-form-dynamic-select/blob/master/gulpfile.js#L61)-section. This way, loading bootstrap-decorator.js correctly causes angular-schema-form.js to load.

However, I am not completely sure about all of this, UMD and ES6 modules and their naming conventions and when what is used is a bit hard to understand and not always clear. So if someone more knowledgeable than me could chime in would be great.

nicklasb:
@angelaharalson, @davidlgj : I feel pretty sure that the correct solution is:
```js
            dependencies: function() {
              return [
                {name: 'schemaForm',
                amd:"angular-schema-form",
                cjs: 'angular-schema-form'},
              ];
```

nicklasb:
Aaand now something seems to be strange in gulp-umd. I'll be back.

PR #496, opened 2015-07-31
Field IDs Patch

plong0:
Updates name and ids (along with corresponding label for="..." attributes) to use the full key (hyphen-delimited) rather than only the last piece.
If a parent FormController is found, we prepend the form's name onto the id values.

We also append the hyphen-delimited key to form.htmlClass, allowing styling of specific fields without having to manually set an htmlClass for each field.


plong0:
looks like tests fail because input names have changed

davidlgj:
Hi @plong0 , I will merge this and #463 when the new builder is up and running. Soonish I hope :)

plong0:
Sounds great @davidlgj!  Thanks, looking forward to it :)

plong0:
made new PR #560 off development branch

PR #521, opened 2015-08-26
Missing control-label class in textarea

vicary:
Consistency on title colours after validation.


Anthropic:
Hi @vicary thank you, as you may know, this has been fixed in the boostrap decorator repo.

Thank you so much for taking the time to submit, look forward to seeing more :+1:

PR #525, opened 2015-08-26
update docs to include sortOptions key on array form objects

devSoss:
This is to update the documentation with the change that was merged into the boostrap decorators repository.


davidlgj:
Awesome!

PR #552, opened 2015-09-21
fixes

DanielSchaffer:
- removed bootstrap-decorator from package definitions
- re-ran gulp (looks like the dists were missing a bunch of changes)


Anthropic:
Hi @DanielSchaffer can you explain what each of the changes are, does this pull relate to some pre-existing issues you can add to your comment?

Also we need the dist file removed, they are only update when a build for release is done.
Also the console log isn't required, the warn seems ok though.

However, as you may be aware the breakout of the core has begun, so it is probably best to wait for that to complete and look at how we integrate any changes with that in mind either in this repo or the core repo instead. So it is up to you if you wish to wait till that completes before doing anything further  on this PR based on my comments :)

DanielSchaffer:
@Anthropic yeah, I'm fine with waiting for the core breakout.

PR #559, opened 2015-09-28
Formselect: fix required and path

gazpachoking:
The path was getting set in the builder as the string transcription of the items array (`[Object, Object]`) or something like that.

Required property would only be set by a parent object, right? So it just makes sense to pass on the one from ourselves to the subforms.

### DIFF
--- a/src/directives/formselect.js
+++ b/src/directives/formselect.js
@@ -12,43 +12,37 @@ angular.module('schemaForm').directive('sfFormSelect', ['sfSelect', 'schemaForm'
         // Keeps the model data for each form
         var formData = [];
         // TODO: Watch the model value and pick a form to match it
-        var once = scope.$parent.$watch(attrs.sfFormSelect, function(form) {
-          if (!form) {
-            return;
+        var form = scope[attrs.sfFormSelect];
+        var model = sfSelect(form.key, scope.model);
+
+        // Watch the model value and change to a form that matches it
+        var key = sfPath.normalize(form.key);
+        scope.$parent.$watch('model' + key, function(value) {
+          model = scope.modelData = value;
+          // If the selected form still validates, make sure we don't change it
+          if (angular.isNumber(scope.selectedForm.value)) {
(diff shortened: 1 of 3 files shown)


gazpachoking:
Added some more fixes in. (probably also requires my pull to decorator branch to work) Things are still a bit weird inside arrays. In the second example the array item gets set to null by default, so when you try to set object properties nothing happens. If the model has an object instead of null already, everything works fine.
The directive that remembers the model value for each form selection also isn't working inside the array example, haven't debugged that one yet. The model also doesn't get cleared when that form is destroyed, probably related.

gazpachoking:
@davidlgj Something I've been wondering for a while: why is the root form restricted to being a fieldset? Are you against lifting that restriction? I just changed so that the array format for fieldsets (like the root form) is allowable in nested forms too (see my 3rd anyOf example,) okay with that?

davidlgj:
@gazpachoking I just merged it and the development branch, haven't started testing it yet though :) 

Regarding "[...] root form restricted to being a fieldset?", I'm not sure I follow what you mean here. But looking at the example you'd like to be able to add an array as an item and have that default to a fieldset? I don't know... if we change it here we must change it for all places that have "items" or in some other way include a subform, i.e.  tabs. I'll have to think about it :) But my gut feeling is not enough bang for the buck... it's not that much more to type and it adds to the complexity of the format. On the other hand we already have shorthands like just the key name `["name"]`.

davidlgj:
One more thing, I think I rather have `[['a', 'b']]` default to a section (that is just a div), since a fieldset usually have  a title, what do you think?

gazpachoking:
> Regarding "[...] root form restricted to being a fieldset?

For example, what if you want a form that returns an array? Or just a simple string? (not a form that returns an object with a value that is an array or string.)

> One more thing, I think I rather have [['a', 'b']] default to a section (that is just a div), since a fieldset usually have a title, what do you think?

Yeah, this is probably okay. I had the impression that an object in the model data should be a fieldset in the form, but I guess that isn't really required.

PR #578, opened 2015-10-17
Ensure key is set in fieldsets

zackbloom:
(also a slight whitespace fix)


Anthropic:
@zackbloom thanks for your contribution sorry it took so long to get reviewed, lots starting to happen :)

PR #583, opened 2015-10-26
introduce main ideas in opening of doco

ErichBSchulz:
as a newbie, this would have help kick start me!


Anthropic:
Thanks for the input Erich

PR #605, opened 2015-11-24
Fixed angular element children selector usage

apinnecke:
In jqLite, `.children()` selector is not able to take an selector argument. Caused by this error, the directive is removing all children, if no full jQuery is used.


apinnecke:
Broken tests also broken in develop. ;)

Anthropic:
@apinnecke thank you for the PR, I'm fairly new to managing the project and we are slowly starting to ramp back up and reviewing some neglected PRs, is there any chance you could write a test script for the change? It would be a huge help :)

apinnecke:
Unfortunately I'm very much out of time these days. Ya, probably could write some tests, but can't tell when that will be. Plus: This change is already covered by the original tests AFAIK, so maybe these are enough? Has to be verified indeed.

nicklasb:
@apinnecke Do you have some time to give some love to this PR?

Anthropic:
@apinnecke thank you for the PR I finally got to it, I can't merge due to the number of changes since but I added the change to sf-schema.directive.js so thank you again for your help.

PR #643, opened 2016-02-24
set firstDigest to false in schemaValidate directive

ruanyl:
In my specific use case, I have a customized multi-select form element. But when the form is first loaded, `scope.firstDigest` of this directive is `false`. AT: [schema-form.js#L2858](https://github.com/json-schema-form/angular-schema-form/blob/development/dist/schema-form.js#L2858)

Which makes it to validate this form element, and update `ngModel.$error` at the first time. This is not the expected behavior I think. `scope.firstDigest` here should be true the first time it is loaded.

I could solve the issue by increasing the timeout at [schema-form.js#L2634](https://github.com/json-schema-form/angular-schema-form/blob/development/dist/schema-form.js#L2634) to for example: `1000`. So, I guess this may related to the sequence how angular initialize the directives and how much time each directive will take.

This pull request do solved my issue, but I'm not sure if there are more elegant ways(or if this breaks anything). feel free to comment :)

### DIFF
--- a/src/directives/newArray.js
+++ b/src/directives/newArray.js
@@ -20,6 +20,7 @@ function(sel, sfPath, schemaForm) {
         // validateField method is exported by schema-validate
         if (scope.ngModel && scope.ngModel.$pristine && scope.firstDigest &&
             (!scope.options || scope.options.validateOnRender !== true)) {
+          scope.firstDigest = false;
           return;
         } else if (scope.validateField) {
           scope.validateField();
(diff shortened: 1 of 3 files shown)


Anthropic:
@ruanyl could you perhaps create a gist or plunker to demonstrate the different behaviours? Help us to see what is happening and what you expect to happen more clearly.

nicklasb:
@ruanyl I agree, it would really help.

Anthropic:
@ruanyl can you check out the issue in the alphas and see if this is still needed in some way or not? Please re-open if it is and we can work on it, but I am hopeful it is no longer an issue :)

Issue #121, opened 2014-10-14
Global options does not propagate properly


davidlgj:
See https://github.com/Textalk/angular-schema-form/issues/29#issuecomment-59048491

nertus:
Any fix?

davidlgj:
Nope not yet.

mån 2 mars 2015 kl 12:16 skrev Aliaksandr Pshanichnik <
notifications@github.com>:

golfmat1:
Would I be correct in assuming that this affects all global options, in particular readonly? E.g. I'm having trouble getting the following to work, i.e. the entire form is still editable:

```html
<div sf-model="..." sf-form="..." sf-schema="..." sf-options="{ formDefaults: { readonly: true } }"></div>
```

golfmat1:
Created small pull request (#396) that fixes my issue for the initial form rendering. 

However, if the model/schema is changed, any updates to the global options aren't realized in the newly rendered form: e.g. if I display one model as readonly then try to show another in the same form as not readonly, it will still not be editable.

davidlgj:
@golfmat1 is that after a `schemaFormRedraw` event?

golfmat1:
I'm not sure. I tried to broadcast that event and still had the issue

However, I'm relatively new to angular and working within a framework in which there are numerous things I could be doing wrong. So it's perfectly possible that I wasn't triggering it correctly.

I created a workaround for now using schemaFormProvider.postProcess to explicitly set readonly on all form elements. This works, though it is probably not ideal.

davidlgj:
@golfmat1 I'm asking since you need to either change the entire form (or schema) instance, otherwise you need the event, and its easy to forget :)

plong0:
any update on this one?

having the same problem described in other threads, formDefaults validationMessage applied to top-level items, but not into sub-sections

jmaister:
I am having the same problem here.

I try to toggle the form between read-only and edit but it does not update the form state.

    $scope.basic.options = { formDefaults: { readonly: true } };
    $scope.toggleEdit = function() {
        $scope.basic.options.formDefaults.readonly = !$scope.basic.options.formDefaults.readonly;
    };

cmartin81:
Any update on this issue?

doingweb:
[...]

PR #644, opened 2016-02-25
Fix for #121 to redraw with proper defaults

stanimoto:
The problem was that initialForm was used to generate a form on redraw
event, but when it was an object type, it was modified in-place by
rendering function. So, when the next redraw event happened, initialForm
already had the properties defined, and it prevented the default values
from being populated into them.


Anthropic:
@stanimoto thanks for the pull request, can you remove the dist folder from the PR, we update that only prior to a release? Also if you are interested in getting involved in the project further we are currently planning the project's future in the Gitter.

ありがとう、
マルセルさん

stanimoto:
Thank you for your response, @Anthropic . I created a new one without the dist folder as  #647 .
(Sorry if it is not the right way to do it since I'm not familiar with PR workflow on GH..)

Anthropic:
I saw it @stanimoto thanks, FYI you can amend a pull request without needing to submit again. But that's fine we just appreciate your efforts. I will confirm the commit or provide further feedback soon :)

PR #647, opened 2016-03-03
Fix for #121 to redraw with proper defaults

stanimoto:
The problem was that initialForm was used to generate a form on redraw
event, but when it was an object type, it was modified in-place by
rendering function. So, when the next redraw event happened, initialForm
already had the properties defined, and it prevented the default values
from being populated into them.


Anthropic:
Satoshi san this is now in release 0.8.13 which is now in bower and the npm release will follow by tomorrow, thanks again!

PR #661, opened 2016-04-06
dist/bootstrap-decorator.js remove

juanmav:
It isn't anymore in this repo


Anthropic:
@juanmav thanks for the PR, but due to following semantic versioning, we technically need to leave the files there until we are ready for our 1.0.0 release.

PR #663, opened 2016-04-11
Add check for schema property to prevent form generation failures whe…

vinceis1337:
…n a checkbox is explicitly defined by a developer (and therefore does not have a schema property)

Fixes: https://github.com/json-schema-form/angular-schema-form/issues/558

### DIFF
--- a/src/services/schema-form.js
+++ b/src/services/schema-form.js
@@ -384,8 +384,14 @@ angular.module('schemaForm').provider('schemaForm',
 
         // Special case: checkbox
         // Since have to ternary state we need a default
-        if (obj.type === 'checkbox' && angular.isUndefined(obj.schema['default'])) {
-          obj.schema['default'] = false;
+        if (obj.type === 'checkbox') {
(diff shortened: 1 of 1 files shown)


Anthropic:
@vinceis1337 looks good, only I forgot to mention that as we are splitting the core out of angular, could you update to use plain javascript instead of the angular.isUndefined?

vinceis1337:
@Anthropic done!

Anthropic:
Thanks, looks good, before I merge I just want to figure out why one test fails, I believe it was **already broken**, but as it says "schemaForm #defaults() should generate default form def from a schema FAILED" I think it should be checked. Probably the test is just out of date.

Thanks for your support @vinceis1337 I'll try to get it sorted soon :)

Anthropic:
I have made the change to the core, it will be in the next alpha, thank you @vinceis1337.

vinceis1337:
@Anthropic thanks for the update, very excited to try out 1.0.0 soon!

Issue #590, opened 2015-11-10
Problems with model binding when using nested Arrays


elbunuelo:
I'm currently using schema form version 0.8.12 with schema form bootstrap version 0.2.0 and I'm trying to use two nested arrays, however the second level array field values end up appearing as [Object object].

I traced back the error to the `keyRedaction` state attribute inside the `array` builder (https://github.com/Textalk/angular-schema-form/blob/0.8.12/src/services/builder.js#L136).

From my point of view, the problem is that the keyRedaction from the parent array is carried over to the child array, which causes the `key` to be an empty array inside the `ngModel` builder (https://github.com/Textalk/angular-schema-form/blob/0.8.12/src/services/builder.js#L31).

By changing line 136 inside builder.js from `state.keyRedaction = state.keyRedaction || 0;` to `state.keyRedaction =  0;` (i.e. resetting the keyRedaction with each nested array), I was able to get it to work properly, but I'm not sure what other repercussions can that change have in the rest of the code.

What do you think about it?

Anthropic:
@elbunuelo I've just started to manage the project and would like to thank you for providing a solution in the issue you have raised, it seems to be working for a lot of people, I will see about getting it in the next update.

elbunuelo:
Well, this makes me really happy. If it's all right, I'd like to submit a pull request for the fix

Anthropic:
By all means :)

mpina83:
finally i will be able to remove the manually override from the plugin xD

Anthropic:
A PR has been merged for the issue, please test the solution is resolved and re-open if it is not.

PR #672, opened 2016-04-26
Proposed fix for issue 590

elbunuelo:
#### Description

Fixes issue 590, which describes a bug that appears when you're using nested arrays. 
The described bug causes element values in the second nesting level to appear as [Object object].

#### Fixes Related issues
https://github.com/json-schema-form/angular-schema-form-bootstrap/issues/9

#### Checklist

Reset keyRedaction variable so that the parent's keyRedaction will not be carried onto its children which used to cause the key to be an empty array inside the ngModel builder.

### DIFF
--- a/src/services/builder.js
+++ b/src/services/builder.js
@@ -133,7 +133,7 @@ angular.module('schemaForm').provider('sfBuilder', ['sfPathProvider', function(s
       var items = args.fieldFrag.querySelector('[schema-form-array-items]');
       if (items) {
         state = angular.copy(args.state);
-        state.keyRedaction = state.keyRedaction || 0;
+        state.keyRedaction = 0;
         state.keyRedaction += args.form.key.length + 1;
 
         // Special case, an array with just one item in it that is not an object.


joelwkent:
Thanks @elbunuelo, I have a test in [PR 652](https://github.com/json-schema-form/angular-schema-form/pull/652) that currently fails due to this issue (#590). So this PR should make my test pass!

elbunuelo:
@joelwkent I'm glad to read it! Now I just have to work out why aren't the tests passing

Anthropic:
@elbunuelo there are some tests failing because of styles not being set, I think David said those tests should be moved to the decorators instead anyway, but I havent had a chance to fix it yet. Nice PR, glad to see the template properly filled in, I will fix it's styling today, nice PR :)

Anthropic:
@joelwkent did you get a chance to see if this resolved your test issue?

joelwkent:
Yes, happy to confirm this PR fixes the issue in the test I have added in the arrayIndices PR.

PR #712, opened 2016-06-10
Custom validation on angular 1.2 fix

montezume:
#### Description

I'm working on a project with Angular 1.2 and was getting an error about $validate is not a function, so I added a little check to resolve the problem. It fixed the issue I was having.

#### Fixes Related issues
haven't found any.

#### Checklist


Anthropic:
@montezume thanks for the PR, do validations continue to work without any call to replace the one removed? I am asking as I don't know how 1.2 validation was working as I started using angular more at 1.3.

I should also add that for version 1.0.0 we will be dropping support for 1.2 as there's too much it doesn't support and the upgrade is fairly painless. Took my team only a day or two to migrate an epic and hideous code base.

But happy to accept if you can confirm behaviour still works and validates in case we release another 0.x.x release.

montezume:
I believe instead we can use $setViewValue(value); This will trigger all of the parsers (the 1.2 equivalent of validators) and rerun the validations. I will try it out and then update my PR. Thanks :)

https://code.angularjs.org/1.2.28/docs/api/ng/type/ngModel.NgModelController

nicklasb:
@montezume Are you still on 1.2, is this important to you? 
@Anthropic Is this check important given we aren't supporting 1.2?

montezume:
I am but I think it's unreasonable to expect 1.2 support in 2016 :) I have just been using a forked version that I have done some 1.2 fixes to

nicklasb:
@montezume The problem is that 1.0 of ASF is a major rewrite of the component that will not be tested at all against 1.2, so it is likely to break that. 
And it is not likely that any large amount of bug fixing will be done in 0.x .versions.

Another problem with the PR itself is that it moves a large chunk of code, which causes issues with diffs over that point. Actually, that is a problem with a specific commit, so perhaps the entire PR has to be redone anyway.

nicklasb:
@montezume My last point was complete and utter nonsense as I didn't see that it was the dist file. I didn't see that you removed it later on. Please disregard. However the first point still stands. :-) 

I do however have a proposition, can't you (for now) maintain a 1.2-1.3 compatible branch of ASF 0.x? (or even 1.x if you wanted to pick off the most important new features)

We could mention it in the ASF docs, that way you could cooperate with others in the same situation and perhaps there could even be a “angular-schema-form-legacy-ng”-package release?

montezume:
A separate branch would make sense I guess, but I'm not sure that there are enough changes required to make it worthwhile.

PR #714, opened 2016-06-13
src: bind scope as ctx of onchange and validators

tflanagan:
####Description

Needed to validate field against value of another field, this seemed the be the only way.

####Checklist
- [ ] I did NOT include the dist folder in my PR


Anthropic:
@tflanagan what is your use case? I've made add-ons for myself before that validate multiple fields together and do calculations, I may be able to help. Regarding this PR I'll check with @joelwkent what he is up to regarding changes to event signatures and see if this change would pass testing.

tflanagan:
@Anthropic Sure no problem:

Validation: If I have Field A as a date field and Field B as another date field, I want to ensure Field B is on or after Field A.

OnChange: If I have Field C as a select and Field D as a date field, I want to auto-fill Field D when Field C is equal to "Foobar".

As the code stands before this PR, I would have to create an addon to do something that, to me at least, should be apart of the framework, which is the ability to interact between fields, updating the model as required.

I may have completely missed how to do this, but from my digging, I couldn't find a way, without writing code.

With this PR, I can have this logic inside the native form properties.

slyadams:
I'm currently writing some onChange handlers which fire whenever any of a set of fields are changed, it calls an http service to generate the value of another field. I believe I'm currently hitting the same limitation addressed there whereby the execution context of the onChange function/expression is quite limited.

nicklasb:
@tflanagan This is a nice feature. However, the authors list, I feel, only currently contain the top contributors, no offence, but..

tflanagan:
@nicklasb take it as it is or don't. We're already using a forked version, so upstream is an afterthought for us.

As for your opinion on the matter of the authors list, that is against every other open-source lib I've encountered and could push potential people away. Anyone who helps, regardless of scale, should be credited in an open-source project.

nicklasb:
@tflanagan I am sorry that you feel that way, but up until now, and AFAIK, we haven't had any demanding to be in the authors list to contribute, people are instead credited on the releases which seems visible enough for most.

Most larger projects doesn't even have a contributors list, which perhaps would be more appropriate given the number of contributors to this project.

raquintero:
@tflanagan Nice feature, I think I need this, but do you have an example of how it is used?

scottux:
I think this use case could best be handled by integrating a JSON rules engine. https://www.npmjs.com/package/json-rules-engine

I saw https://www.npmjs.com/package/react-jsonschema-form-conditionals when I was first looking to do something similar, and thought it would be a good approach when using this library rather than that react library.

PR #729, opened 2016-06-27
fix package.json main file

mathroc:
When I tried to include 'angular-schema-form' (`import schemaForm from 'angular-schema-form';`) it would produce this error when trying to "compile" it with webpack:

> Module not found: Error: Cannot resolve module 'angular-schema-form/schema-form' in /scripts/src

It can be work-around like this until this is fixed here:

`import schemaForm from 'angular-schema-form/dist/schema-form';`

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0-alpha.2",
   "description": "Create complex forms from a JSON schema with angular.",
   "repository": "json-schema-form/angular-schema-form",
-  "main": "dist/schema-form.es5.js",
+  "main": "dist/schema-form.js",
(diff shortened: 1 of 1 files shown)


mathroc:
in fact for workflows with es6 transpilers, I wonder if it would not be best to provide a index.js with the appropriate imports (angular, tv4, angular and src/*)

and in any case the required runtime dependancies should be added as "dependencies" instead of "devDependencies"

thoughts ?

Anthropic:
@mathroc the devDependencies should be compiled into the final output while the dependencies are not, which dependencies are not included? json-schema-form-core is embedded by babel so it is not required on a prod system independently. Don't believe anything else is required.

PR #748, opened 2016-08-02
form '*' generate only remain keys from schema

illiano:
#### Description

Add your description here
#### Fixes Related issues
- add related
- issues here
#### Checklist

- it is now possible to add custom keys to the form and add '...' for the remaining keys

### DIFF
--- a/src/services/schema-form.js
+++ b/src/services/schema-form.js
@@ -313,8 +313,25 @@ angular.module('schemaForm').provider('schemaForm',
       //simple case, we have a "*", just put the stdForm there
       var idx = form.indexOf('*');
       if (idx !== -1) {
-        form  = form.slice(0, idx)
-                    .concat(stdForm.form)
+        var formKeys = form.map(function(obj) {
+          if (obj.key) {
+            return obj.key;
+          }
+        }).filter(function(element) {
+          return element !== undefined;
+        });
+
+        var stdFormRemains = stdForm.form.map(function(obj) {
+          isInside = formKeys.indexOf(obj.key[0]) !== -1;
+          if (!isInside) {
+            return obj;
+          }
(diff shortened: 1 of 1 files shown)


Anthropic:
@illiano thanks for taking the time to make a PR I really appreciate it!
I was wondering if perhaps '...' may be a better identifier for 'rest' so the * can remain as it is, would that be a hard change to make to your PR?

Anthropic:
@illiano thanks for the update, there's two things now that would get this merged quicker:

1) documentation update, I would have to find time to update the documentation before merging, if you can help with that I'd appreciate it, just needs to spell out behaviour so a user knows that it will only add fields not already added at the order position of where the '...' is added (assuming that's what it does?)

2) if you could make a plunker to show it in action?
Without a test case it would be nice to be able to ensure it is stable without having to do that myself, would really appreciate your time if you can help with that :)

Otherwise I will do my best to get this merged soon :)

illiano:
I have created a plunker, where you can see the changes in action. 

https://plnkr.co/edit/7Oqxxl?p=preview

I think it's better, when you update the documentation :-)

Anthropic:
@illiano I'm not sure the plunker is working, I changed the form to be:
```
  $scope.form = [
    "name",
    "...",
    "comment"
  ];
```
And the form renders 
"name",
"name",
"email",
"comment",
"comment"

Which is the same output as just using "*", is something missing?

illiano:
i have fixed the bug.

Anthropic:
Great, thank you very much @illiano I will merge once I make the same changes to the webpack branch hopefully I will have time tonight, then I will look at adding documentation for it.

Anthropic:
@illiano this is in the alpha :)

Issue #754, opened 2016-08-05
Change title dynamically 


milansgowda:
#### Question

As a user  I should be able to change the title/dynamically of the form dynamically. Currently i can change the value of inputs, is there a way where i can change title/label dynamically
#### Expected behaviour

{
    "key": "comment",
    "type": "textarea",
   "title":"user.title",
    "placeholder": "Make a comment"
  }
where i can set value for comment like $scope.model.comment ="Comment description" and it would change the value.  Same set $scope.user.title = "First Name" it should display "First Name" as title
#### Actual behaviour

It actually showing user.title as title

@json-schema-form/angular-schema-form-lead

MGouaillier:
Hello :)
See : #538 

Hope this can help

Anthropic:
Using the new builder this may be possible without the overhead added by two new watchers per field if the suggested PR doesn't cover what you are after.

Anthropic:
Please re-open if you do not feel the question has been answered sufficiently I'm spring cleaning quiet issues.

PR #755, opened 2016-08-08
Add support for function instead of string in name/description field …

fookoo:
#### Description

Add support for function instead of string in schema object.

You can now use 

```
let schema = {
        type: "object",
        properties: {
                name: {
                    type: "string",
                    minLength: 2,
                    title: () => $filter('translate')("Name"),
                    description: () => $filter('translate')("Description")
                },
                title: {
                    type: "string",
                    enum: ['dr','jr','sir','mrs','mr','NaN','dj']
                }
            };
```
#### Fixes Related issues
#754
#### Checklist

### DIFF
--- a/src/services/schema-form.js
+++ b/src/services/schema-form.js
@@ -70,12 +70,26 @@ angular.module('schemaForm').provider('schemaForm',
     var f = options.global && options.global.formDefaults ?
             angular.copy(options.global.formDefaults) : {};
     if (options.global && options.global.supressPropertyTitles === true) {
-      f.title = schema.title;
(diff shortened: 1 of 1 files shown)


Anthropic:
@fookoo thanks for the PR, can you see a way we could apply the solution to the form instead of the schema? schema-form is a standard we do not control and can't really make changes to while remaining compliant.

I love the approach and can see the need for it, but to be accepted it needs to be applied via the form javascript instead which is optional also.

Anthropic:
Closing due to lack of response, i18n has alternative suggestions for handling such as #865

PR #759, opened 2016-08-17
Update index.md

lucasgranberg:
####Description
Small documentation fix

"validationMessages" should be "validationMessage" and as it is a object the keys need values.


lucasgranberg:
Nevermind. It is already merged here:
https://github.com/json-schema-form/angular-schema-form/pull/469

PR #809, opened 2016-12-02
adding raw template to decorator

Traksewt:
####  Description

Currently setting the template for a custom decorator requires using the template cache, and you cannot pass in a raw template string. There is some code inside the decorator provider to support it, but it is not exposed, as it does not work.

When manually setting the decorator to have type equals to template, the builder cannot find the raw template. This is because it is looking for the 'template' field on the wrong variable, checking the form instead of the 'field'.

To test out using a raw template, you can try
```
schemaFormDecoratorsProvider.decorator('bootstrapDecorator')['myCustomDecorator'].type = 'template';
```

####  Fixes Related issues
- This doesn't expose raw templates to the decorator helper methods, however, it does get rid of the bugs so it can be exposed in the future.

####  Checklist
- [Y] I have read and understand the CONTRIBUTIONS.md file
- [Y] I have searched for and linked related issues
- [Y ] I have created test cases to ensure quick resolution of the PR is easier
- [Y] I am NOT targeting main branch
- [Y] I did NOT include the dist folder in my PR

### DIFF
--- a/src/services/builder.js
+++ b/src/services/builder.js
@@ -277,8 +277,8 @@ angular.module('schemaForm').provider('sfBuilder', ['sfPathProvider', function(s
        */
       build: function(form, decorator, slots, lookup) {
         return build(form, decorator, function(form, field) {
-          if (form.type === 'template') {
-            return form.template;
+          if (field.type === 'template') {
+            return field.template;
           }
           return $templateCache.get(field.template);
         }, slots, undefined, undefined, lookup);
--- a/test/services/decorators-test.js
+++ b/test/services/decorators-test.js
@@ -26,4 +26,28 @@ describe('schemaFormDecorators', function() {
       });
     });
   });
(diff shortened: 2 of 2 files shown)


Anthropic:
@Traksewt thanks for your PR sorry for the delayed response, I was in the middle of the separated core changes and then holidays, would you be able to update to the latest version and reapply your changes to the renamed files? Would love to accept your efforts :)

src/services/builder.js -> src/services/sf-builder.provider.js
test/services/decorators-test.js -> src/services/schema-form-decorators.provider.spec.js

Traksewt:
@Anthropic ok no problem. rebased.

Anthropic:
Didn't realise I could change it myself within your commit, how cool :)

Anthropic:
@Traksewt any ideas?

scottux:
The tests are failing because the decorator is trying to load the template remotely as a url, but the template is a string.

PR #824, opened 2017-01-17
changed schema-form.js to angular-schema-form.js in builder examples

CatherineH:
This is a very small update to the examples.
I believe the angular-schema-form.js filename was changed.


Anthropic:
Thanks @CatherineH I forgot about those two :)

Issue #870, opened 2017-04-25
form.key inside arrays


mnzaki:
#### Bug
`form.key` inside arrays appends the array path and empty string `""` after the actual key, for example giving `["names", 0, "names", ""]` for the first array item in an array with key `["names"]`.

#### Gist/Plunker/Demo
[Plunker demonstrating the issue](https://plnkr.co/edit/e7gj799X4ijPyUEbyA0n?p=preview)

@json-schema-form/angular-schema-form-lead

mnzaki:
And it has more weird behavior when in arrays in objects. Plunker updated to demonstrate.

Anthropic:
@mnzaki `"key": 'names[]'` should be `"key": 'names'`
You only need [] when there is a sub property like `names[].names`

Because it doesn't know the key it can't replace it with the index.

mnzaki:
I updated the plunker to reflect that, but now the empty string gets replaced with `0` (for all keys), also `form.key` seems to be overridden by the last key always (try adding a new value to the array)

mnzaki:
Also the behavior of `sfKeyController` seems to be broken for `type: "fieldset"`. The `fieldset` form addon adds an `sf-index="{{$index}}"` attribute, but there's no `$index` defined in that scope and that gets parsed into a `0`.

Anthropic:
@mnzaki

mnzaki:
I have been looking at code, and there seems to be a few problems.
For starters, `sfPath.parse("[]")` returns `[""]`, is this correct behavior?

If so then in `fieldset.html` addon template, `sf-key-controller sf-parent-key=\"[{{form.key.join('][')}}]\"  sf-index=\"{{$index}}\">` will produce incorrect results when `form.key` is empty, as that evaluates to `sf-parent-key="[]"` which is parsed to `[""]` and then the index (which is empty) is parsed to `0`.

Anthropic:
@mnzaki right, makes sense, I didn't get a chance the other night, wife dragged me out shopping, very much against my will, will definitely take a look this weekend.

mnzaki:
More analysis:

Another related problem is in [this line](https://github.com/json-schema-form/angular-schema-form/blob/104a3fa98d688b547feece7d4a38bc7c20adcf5e/src/directives/sf-field.directive.js#L34) in `sf-field.directive`:
```javascript
        scope.initialForm = Object.assign({}, sfSchema.lookup['f' + attrs.sfField]);
[...]

PR #874, opened 2017-04-29
Fix form.key issues

mnzaki:
####  Description
Generate proper form keys for arrays of strings and arrays of objects, and add some tests for it.
This also depends on a fix in angular-schema-form-bootstrap, specifically removing the `sf-key-controller` from `fieldset.html`

####  Fixes Related issues
Issue #870 

####  Checklist


mnzaki:
Note that tests will fail because this depends on json-schema-form/angular-schema-form-bootstrap#42

Anthropic:
@mnzaki I added your bootstrap change and it is still failing on strings, is something else missing?
    ✔ should generate correct form keys for array of objects
    ✔ should generate correct form keys for array of objects with user specified form
    ✖ should generate correct form keys for array of strings

Oh and thanks for noticing the dupe of tests, they were previously breaking the loop so I had to split them out to update the loop logic and forgot to clean them up.

mnzaki:
It needs to be built as the karma tests use the built file in `dist/`, and I did not include that in the PR as required. The "array of objects" tests pass as they do not need changes in angular-schema-form, they only require the bootstrap fix.

mnzaki:
Also there's one more fix missing here, the issue that I explained in [this comment](https://github.com/json-schema-form/angular-schema-form/issues/870#issuecomment-297963137):
```
        scope.initialForm = Object.assign({}, sfSchema.lookup['f' + attrs.sfField]);
/*->*/  scope.form = sfSchema.lookup['f' + attrs.sfField];
```

I was hesitant about fixing this as well as it might introduce bugs in projects using ASF and depending on this unexpected side effect (ie, updating form object and getting things changed without broadcasting a redraw event). What do you think?

Anthropic:
@mnzaki I did the build, that's why it started passing two of the three, it was failing all three originally :)

mnzaki:
I don't really fully know how the tests are run on Travis. But on `formKeyFixes` branch the angular-schema-form build inside `dist` is not updated, if you make a clean checkout and run the tests they will fail, but if you `npm run build` the tests will pass.

Anthropic:
@mnzaki travisci clones the whole repo, then apply and merge the pr before running the tests, you can see the output here https://travis-ci.org/json-schema-form/angular-schema-form/pull_requests I believe that's visible to anyone. You can see that after I committed the re-built dist folder the test result changed, objects and strings failed in the last run but then only strings the next time. I had a quick look last night, but I need more time tonight to hopefully fix the strings, seems like there just a duplication of the key I need to eliminate.

mnzaki:
@Anthropic yes and then it runs the tests through `npm run test` which uses `dist/angular-schema-form.js` (not from `src/`) so it needs to be built and committed as I did just now, then tests pass.

Anthropic:
@mnzaki ah d'oh sorry was in too much of a rush to notice, I really should update the tests to run a build into a folder ignored by git... one day... thanks for taking the time to make sense for me! :)

I will try to look at your other suggestion shortly.

mnzaki:
@Anthropic btw I think just changing the npm test script to `"webpack && karma start karma.conf.js --log-level debug --single-run"` would be enough to run the travis tests properly.

Anthropic:
@mnzaki cheers, updated locally, will push on the weekend :)

PR #895, opened 2017-06-13
Fix array item deletion issue

mnzaki:
####  Description
Deleting array items which are objects will cause the fieldset field's `sf-field` directive `$destroy` event handler to try to remove itself from the array again. This was broken in various way previously because of incorrect form keys, but the issue became visible after changes from #870 

####  Checklist


mnzaki:
Note failing test because travis still not running the build before the tests

Issue #900, opened 2017-06-19
schemaForm.merge reports incorrect keys for array object properties


donalmurtagh:
## Overview

In the latest version (1.0.0-alpha.5), the `schemaForm.merge` function reports incorrect keys for object properties stored in an array

## Steps to Reproduce

1. Open [this Plunker demo](https://plnkr.co/edit/H1ZEhzr5J9yRNouR5vVN?p=preview)
2. Click the "Merge form and schema button" and look at the result
3. In the result section, look at the keys reported for the child's name and age

### Name
```
                "key": [
                  "",
                  0,
                  "peopleLivingWithYou",
                  "dependentChildren",
                  0
                ],
```
### Age
```
                "key": [
                  "",
                  0,
                  "peopleLivingWithYou",
                  "dependentChildren",
                  0
                ],
```

Obviously these properties should not have the same key, and `name` and `age` should be last elements of the `key` array.

If you remove the `section` from the form, i.e. change the form definition to

```javascript
    $scope.form = [{
        key: 'peopleLivingWithYou.dependentChildren',
        add: "Add Child",
        title: 'Dependent children details',
        validationMessage: 'Complete all required fields for at least one child'
      }];
```

The keys are correctly reported as

### Name
```
            "key": [
              "peopleLivingWithYou",
              "dependentChildren",
              0,
              "name"
            ],
```

### Age
```
            "key": [
              "peopleLivingWithYou",
              "dependentChildren",
              0,
              "age"
            ],
```

This problem does not exist in version 0.8.13.

@json-schema-form/angular-schema-form-lead

Anthropic:
Thanks @donalmurtagh I will turn this into a test case to ensure the scenario is supported in future changes as well.

donalmurtagh:
@Anthropic I discovered that you can workaround the problem by using `fieldset` in the form definition instead of a `section`. You can verify this in the Plunker example

Anthropic:
@donalmurtagh this test is passing in the latest changes I merged, please re-open if there's anything still not working in relation to this :)

donalmurtagh:
@Anthropic I've updated the Plunker demo to use the latest code and agree that this issue is now resolved

PR #902, opened 2017-06-20
add regression tests for #900

donalmurtagh:
####  Description
This PR adds regression tests for #900. Curiously, the test does _not_ reproduce the issue demonstrated by [the Plunker example](https://plnkr.co/edit/H1ZEhzr5J9yRNouR5vVN?p=preview) even though the schema and form definitions are identical in both.

In the Plunker demo, the keys are:
- name: `["", 0, "peopleLivingWithYou", "dependentChildren", 0]`
- age: `["", 0, "peopleLivingWithYou", "dependentChildren", 0]`

Whereas in the unit tests, the keys are:
- name: `["peopleLivingWithYou", "dependentChildren", "", "name"]`
- age: `["peopleLivingWithYou", "dependentChildren", "", "age"]`

In my particular case, the only thing I really care about is that the last element of the key array is the property name, so if I could get the same behaviour in the Plunker demo (and my app), I'd be satisfied.

####  Checklist

### DIFF
--- a/src/services/schema-form.provider.spec.js
+++ b/src/services/schema-form.provider.spec.js
@@ -617,8 +617,8 @@ describe('schema-form.provider.js', function() {
         var nameKey = objectPropertyKeys.items[0].key;
         var ageKey = objectPropertyKeys.items[1].key;
 
-        nameKey.pop().should.eq("name");
-        ageKey.pop().should.eq("age");
+        nameKey.join('.').should.eq("peopleLivingWithYou.dependentChildren..name");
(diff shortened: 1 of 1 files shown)


Anthropic:
Travis isn't running the tests due to a false positive on malicious content, I've emailed them to get it fixed.

Aside from that I will accept this as soon as I get a chance to work on it, probably tonight, don't want it to show up everything as broken until it is fixed, I've always hated that there's no todo list items in tests so you can add tests that knowingly fail without failing the whole test.

Issue #651, opened 2016-03-17
Bootstrap-decorator is broken with webpack


xbill82:
Hi,
when trying to `require('angular-schema-form/dist/bootstrap-decorator')`, we get the following error

     ERROR in ./~/angular-schema-form/dist/bootstrap-decorator.js
      Module not found: Error: Cannot resolve module 'schemaForm' in /var/app/node_modules/angular-schema-form/dist

Anthropic:
@xbill82 the decorator has recently been split out into its own repo, which versions are you using?

Anthropic:
Most likely the files are not loaded in the correct order, if you can't sort it out please create a plunker to demonstrate what you are doing and we can try to help.

Please re-open if you do not feel the question has been answered sufficiently.

GarrisonKubis:
Hello,

I'm experiencing the same issue. 

```
infra = angular.module 'myapp', [
  require 'angular/angular'
  require 'angular-sanitize/angular-sanitize'
  require 'tv4/tv4'
  require 'objectpath/lib/ObjectPath'
  require 'angular-schema-form/dist/schema-form'
  require 'angular-schema-form/dist/bootstrap-decorator'
]
```

```
Error: Cannot find module 'schemaForm' from '<my_path>/bower_components/angular-schema-form/dist'
```

I am using version [0.8.13](https://github.com/json-schema-form/angular-schema-form/releases/tag/0.8.13) so I think bootstrap-decorator is still in the same repo (even though there is a `WHERE _IS_BOOTSTRAP_DECORATOR.md` in there).

Anthropic:
@GarrisonKubis use the version of the bootstrap decorator from the bootstrap repo for 0.8.13

aelamrani:
+1

legend80s:
+1 The solution: Use the npm package [angular-schema-form-bootstrap](https://www.npmjs.com/package/angular-schema-form-bootstrap).

aelamrani:
@stevenmccord we had to copy files locally to get it work... a bit ugly but when we faced the issue, it was the only solution we found

You can find here how: https://github.com/gravitee-io/gravitee-management-webui/blob/master/src/management/management.module.ts#L77

stevenmccord:
[...]

PR #932, opened 2017-10-05
Add webpack loading instructions

stevenmccord:
####  Description

Add webpack loading instructions to the README.

#651 

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,38 @@ Don't forget to load the `schemaForm` module or nothing will happen.
 angular.module('myModule', ['schemaForm']);
 ```
 
+### Webpack loading
+In order to import angular-schema-form into your application you will need to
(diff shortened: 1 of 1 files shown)


Anthropic:
Thank you very much @stevenmccord I appreciate this, just wanted to check if you have tried and know already if the alpha (develop) is any different or would the same instructions apply? I suspect they would.

stevenmccord:
I have not tried the Alpha, but I know with the bootstrap split out it is different instructions.  I imagine with the alpha I wouldn't need to utilize the import-loader as well so it should be simply `import angular-schema-form` and `import angular-schema-form-bootstrap`, but I can't say for sure yet, although I would like to give alpha a shot we are using the old version in production and can't quite move there yet.  Thanks!

Issue #888, opened 2017-06-02
Array validation broken in Angular 1.5.10


donalmurtagh:
## Steps to reproduce

- Open the [Plunker demo](https://plnkr.co/edit/1fphMFbKXhK5CaAfmntD?p=preview)
- Answer "Yes" to the first question, this will reveal a savings and investments section that allows one or more account details to be entered
- **Problem no.1**: Notice that the "Enter details of at least one account" validation message is shown, even though these fields are pristine and I've specified that errors should not be shown for such fields via
```
sf-options="{pristine: {errors: false, success: false}}"
```
- Complete all account fields
- **Problem no.2**: Notice that the validation message remains even when though all account fields are valid
- Oddly, if you show and hide the savings and investments section by changing the answer to the first question to "No", then back to "Yes", the validation message disappears

**If you change the Angular dependencies to 1.5.9, these problems do not occur**

@json-schema-form/angular-schema-form-lead

Anthropic:
Nice, a breaking change in a patch release, gotta love that. Hopefully that means it is a minor change I can manage a fix for. Thanks for finding the offending release and the clear plunks.

ivarprudnikov:
Simplified Plunker version using unpkg urls: https://plnkr.co/edit/tWUx9bjwok2rrYbSeOjq

ivarprudnikov:
Diff between AngularJS 1.5.9 and 1.5.10
https://github.com/angular/angular.js/compare/v1.5.9...v1.5.10

PR #998, opened 2019-10-25
fix for #888 - array validation broken in angular 1.5.10

alambir:
####  Description
This fixes issue #888. On angular 1.5.10+, `firstDigest` value is `false` on the initial load which triggers validation on empty fields. There might be some changes in the digest cycle/processing between angular v1.5.9 and v1.5.10+

`firstDigest` is initialised in  `sf-schema-directive.js `
```
        // We need to know if we're in the first digest looping
        // I.e. just rendered the form so we know not to validate
        // empty fields.
        childScope.firstDigest = true;
        // We use a ordinary timeout since we don't need a digest after this.
        setTimeout(function() {
          childScope.firstDigest = false;
          scope.$apply();
        }, 0);
```
and consumed in `sf-array-directive.js` in the following block:
```
      // We need to have a ngModel to hook into validation. It doesn't really play well with
      // arrays though so we both need to trigger validation and onChange.
      // So we watch the value as well. But watching an array can be tricky. We wan't to know
      // when it changes so we can validate,
      let watchFn = function() {
        // scope.modelArray = modelArray;
        scope.modelArray = scope.$eval(attrs.sfNewArray);
        // validateField method is exported by schema-validate
        if (scope.ngModel && scope.ngModel.$pristine && scope.firstDigest &&
            (!scope.options || scope.options.validateOnRender !== true)) {
          return;
        }
        else if (scope.validateField) {
          scope.validateField();
        }
      };
```
Tests were carried on angular 1.5.9 and 1.5.10 to check the `$pristine` value of the field (`schema-validatate-directive.js`):
```
        ngModel.$formatters.push(function(val) {
        // When a form first loads this will be called for each field.
        // we usually don't want that.
        if (ngModel.$pristine && first &&
            (!scope.options || scope.options.validateOnRender !== true)) {
          first = false; 
          return val;
        }
        validate(ngModel.$modelValue);
        return val;
      });
```

**Angular 1.5.9**
```
initial load page:
ngModel.$pristine true

click on yes radio button:
ngModel.$pristine true
ngModel.$pristine true
ngModel.$pristine true
ngModel.$pristine true
```

**Angular 1.5.10**
```
initial load page:
ngModel.$pristine true

click on yes radio button:
ngModel.$pristine true
ngModel.$pristine true
ngModel.$pristine false <----
ngModel.$pristine true
```

Removing `scope.firstDigest` from the condition in (`watchFn function - sf-array-directive.js`) `line 22` fixes this issue.  

####  Checklist


ivarprudnikov:
Diff between AngularJS 1.5.9 and 1.5.10
https://github.com/angular/angular.js/compare/v1.5.9...v1.5.10

DH30124 (approved):

john-latham:
This PR appears to include private code changes. It should not be merged.

